PR #487, opened 2018-07-20
Sync: Add support for integrations via webhooks and a integrations user

sqweelygig:
- Added `action-emit-webhook` card and code
- Added `user-integrations` card and role

Change-type: patch


sqweelygig (commented):
This would be "request changes", but I can't request changes of my own PR!

jviotti (changes requested):

PR #493, opened 2018-07-30
Sync: Add user-sync

sqweelygig:
Add user card for sync user

Copy view-read-user-community (as a jumping off point) and add the
single extra action of action-emit-webhook

Add the sync user by bundling the existing code for the password-less
guest user, and invoking it on both guest and sync

Change-type: patch

### DIFF
--- a/default-cards/contrib/user-actions.json
+++ b/default-cards/contrib/user-actions.json
@@ -1,0 +1,13 @@
+{
+  "slug": "user-actions",
+  "type": "user",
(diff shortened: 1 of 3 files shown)


jviotti (changes requested):

jviotti (changes requested):
Are we sure that the actions user needs all the permissions set in the role you created? Can you try removing them and keep only what's needed?

cc @LucianBuzzo

PR #586, opened 2018-08-20
Sync: Bare bones module environment

sqweelygig:
Created a lib/sync sub-folder with its own module structure. Extended
existing .gitignore to include lib/sync built files. Created a
package.json with some basic details. Created a non-entity of a index.ts
to smoke-test compilation. Created a tsconfig.json that takes the repo
default, and specifies that compilation happens in-place. Created a
tslint.json with resin.io rules.

Change-type: patch


LucianBuzzo (changes requested):

LucianBuzzo (dismissed):

jviotti (changes requested):
I left some comments about the directory structure and some other stuff.

jviotti (changes requested):
I left one small comment, but this is looking really good.

Also, we should add `lib/action-library` as a dependency in the top level `package.json`

jviotti (changes requested):
I left some minor comments, but looks much better :+1:

jviotti (changes requested):
I saw these same changes in another PR. Which one should we review? Both Lucian and I left some comments on a very similar PR, so I guess they will need to be fixed here as well?

jviotti (approved):

PR #622, opened 2018-08-24
fix(SDK): Fix bug where error messages would be "[Object object]"

LucianBuzzo:
Fixes #618

Change-type: patch


sqweelygig (approved):
Approving, because it's good as is; but I think it could be better.

PR #648, opened 2018-09-03
SDK: Remove SDK build step as it is only used by the UI

LucianBuzzo:
Change-type: patch


jviotti (approved):
Left some minor comments, but looks good

PR #687, opened 2018-09-13
CI: Remove tests steps that aren't needed

LucianBuzzo:
Fixes #681

Change-type: patch


jviotti (approved):
Isn't this removing the whole cache though?

PR #761, opened 2018-10-09
links: Evaluate links for querys that get an element by slug or id

LucianBuzzo:
Change-type: patch


jviotti (approved):
I left a small note about a test case title, but otherwise it looks great!

PR #888, opened 2018-11-08
worker: Fix expansion concurrency issues on triggered actions

jviotti:
We have a mutable `matchCard` variable that is either a reference to the
card on which we're processing triggered actions, or a deep clone of it
with the `target` property expanded.

Right below its declaration, we're processing triggered actions
concurrently (with `Bluebird.map`), which means that there is a
potential race condition where the `matchCard` variable will be
re-mutated to its non-expanded case after a mutation caused by another
trigger, and therefore a certain triggered action will not run.

Change-type: patch


LucianBuzzo (approved):
🎊

PR #916, opened 2018-11-14
Add redis cache

nazrhom:
Replace in memory-cache with redis. Cards will be cached for 1 hour
before expiring.

Change-type: minor
Signed-off-by: Giovanni Garufi <giovanni@resin.io>


jviotti (approved):
I left some minor comments but overall looks great!

PR #930, opened 2018-11-16
Assert create event

nazrhom:


LucianBuzzo (changes requested):

jviotti (approved):
Looks great! I added a small comment about writing a test case that checks that we can create an event with tag, but otherwise its cool.

LucianBuzzo (approved):

PR #979, opened 2018-11-29
Use global katapult in deploy script

afitzek:
Change-type: patch
Signed-off-by: Andreas Fitzek <andreas@resin.io>


mikesimos (approved):
we can remove the wget line

PR #998, opened 2018-12-04
core: Store type information in link cards

jviotti:
Change-type: minor


afitzek (approved):
LGTM, although I'm not super confident in my knowledge about some parts of JF that are touched here yet.

LucianBuzzo (approved):

PR #1117, opened 2018-12-18
Make a rethinkdb cluster in k8s

afitzek:
Change-type: minor
Signed-off-by: Andreas Fitzek <andreas@resin.io>

Depends-On: https://github.com/balena-io/balena-rethinkdb/pull/1


mikesimos (commented):
We can probably close this one since we're moving away from rethinkdb. Alternatively, we could render the templates of the official rethinkdb helm chart.

PR #1164, opened 2018-12-21
queue: Use a predictable execute event slug based on the request

jviotti:
This means that we don't have to do a full blown unoptimized query, but
just a constant get by slug operation.

Change-type: patch


afitzek (approved):
LGTM

LucianBuzzo (approved):

PR #1311, opened 2019-01-17
server: Attempt to create a card when /ping is requested

jviotti:
So that the health check can discover worker and/or database problems.

Change-type: patch


LucianBuzzo (approved):

mikesimos (approved):
Nice!

PR #1478, opened 2019-02-26
deploy-templates: Setup deploy templates for Postgres

jviotti:
Change-type: minor


mikesimos (approved):
@jviotti good to merge

PR #1539, opened 2019-03-13
deploy-templates: Test deploying 2 HTTP servers

jviotti:
Change-type: minor


mikesimos (approved):
cool :+1: 
Just noting here, that the default session affinity is set to random, but let's see how it goes.

PR #1732, opened 2019-04-10
deploy-templates: Add POD_NAME envvar; Use it in SERVER id

mikesimos:
Change-type: patch
Signed-off-by: Michael Angelos Simos <michalis@balena.io>


jviotti (approved):
Looks great :tada:

PR #2192, opened 2019-07-14
queue: Re-implement using RabbitMQ 

jviotti:
The benefits are:

- A lot less code, as RabbitMQ supports the exact same way our queue
works like
- A decent speed boost
- A lot less queries against PostgreSQL, allowing us to have much more
workers with little overhead

Change-type: major


mikesimos (approved):
deploy-templates LGTM

PR #2321, opened 2019-08-09
Infinite list

karaxuna:
This can be reused in more places when reverse list is implemented (for example for timeline)


LucianBuzzo (changes requested):
Nice abstraction :+1:
If you can separate out your changes from the chat-widget work we can merge this in

LucianBuzzo (approved):

PR #2332, opened 2019-08-13
Add S3 related env vars to deployment template

LucianBuzzo:
Change-type: patch


mikesimos (commented):
Heya @LucianBuzzo  We need the definition of those variables in https://github.com/balena-io/jellyfish/blob/v19.20.4/deploy-templates/staging/v1.0.0/kubernetes/config-manifest.yml as well.

mikesimos (approved):

mikesimos (approved):

PR #2422, opened 2019-08-28
deploy-templates: Enable all necessary RabbitMQ ports

jviotti:
See https://stackoverflow.com/a/12811593

Change-type: minor


mikesimos (approved):
@jviotti  Those ports aren't necesairy in our single node RabbitMQ instance, but may be useful to expose within the cluster/namespace here. Approving.

PR #2479, opened 2019-09-04
Fix make dev-% commands

karaxuna:
Change-type: patch


jviotti (approved):
Good catch. I should have tested this more.

PR #2518, opened 2019-09-19
Add flowdock sync test events

mbalamat:
***

test suite!


jviotti (changes requested):
This looks good and its good to merge (you can then work on the `expected.json` files as a separate PR), but please rebase it from master, and you will notice that all the webhook JSON events now live in `test/integrations/sync/webhooks` instead of `test/e2e/sync/webhooks`. I'll approve after the files are moved to the right place!

jviotti (approved):

PR #2613, opened 2019-10-18
Added support to triggers to work on multiple cards

ffissore:
Field `card` can now be a json-e template that returns an array of card ids

For example
```js
card: {
  $eval: 'source.data.cards'
}
```
when `source.data.cards` is an array of card ids, or
```js
card: {
  $map: {
    $eval: 'source.data.cards[0:]'
  },
  'each(card)': {
    $eval: 'card.id'
  }
}
```
when `source.data.cards` is an array of objects like `{ id: 123 }`

See https://taskcluster.github.io/json-e/

Connects-to: #2547
Change-type: patch
Signed-off-by: Federico Fissore <federico@fissore.org>

### DIFF
--- a/apps/server/default-cards/contrib/triggered-action-support-closed-issue-reopen.json
+++ b/apps/server/default-cards/contrib/triggered-action-support-closed-issue-reopen.json
@@ -58,7 +58,7 @@
     "action": "action-update-card",
     "target": {
       "$map": {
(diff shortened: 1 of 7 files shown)


jviotti (changes requested):

jviotti (approved):
Looks really good! I love the tests :+1:

LucianBuzzo (approved):

PR #2652, opened 2019-10-24
Update deploy-templates for katapult v2

nazrhom:
Change-type: patch
Signed-off-by: Giovanni Garufi <giovanni@balena.io>

***

test suite!


richbayliss (approved):
We could drop the comments in the tpl file above, but it's non-critical

PR #2687, opened 2019-10-30
GitHub integration extend

mbalamat:
***

test suite!


jviotti (approved):
Looks good. @mbalamat Can you, after merging this, open another PR that starts using these test cases (so the PR will initially be red) and then @nazrhom and you can add commits on top to make it pass?

PR #2688, opened 2019-10-30
tests: add polledTest for dynamic test timing

imrehg:
Some tests require some wait before they succeed. A fixed sleep is previously used, but that doesn't allow for faster or slower test environments (such as the connected PR, running tests in balenaOS). Add a helper to create a more dynamic testing setup for such cases.

Connects-to: #2650
Change-type: patch
Signed-off-by: Gergely Imreh <gergely@balena.io>

***

test suite!


CameronDiver (approved):
LGTM

jviotti (changes requested):
cc @LucianBuzzo As you maintain these Puppeteer tests

PR #2704, opened 2019-11-04
server: Correctly handle S3 file retrieval exceptions

jviotti:
Change-type: patch

***

test suite!


ffissore (approved):

ffissore (approved):
Code style wise, I’d put the catch on a new line, as the then above it

PR #2746, opened 2019-11-11
UI: Add a new custom route for sent/received messages

LucianBuzzo:
Fixes #2658

Change-type: minor

***

test suite!


karaxuna (changes requested):
Left some comments about performance optimisations, but they may not have critical impact.

karaxuna (approved):

PR #2769, opened 2019-11-15
Stop showing changelog in a modal

LucianBuzzo:
Displaying changelog items to end users is not meaningful as many of the
changes don't have any bearing on users of the UI.

Change-type: minor

***

test suite!


ffissore (approved):
Totally approve. Should we want to restore it, we should deliver a changelog targeted to end users

PR #2797, opened 2019-11-21
Improve getActor performance

karaxuna:
Change-type: minor

***

test suite!


LucianBuzzo (approved):
Nice!

PR #2800, opened 2019-11-22
Improvements to how sql is generated from json schema

ffissore:
@jviotti @LucianBuzzo please give these commits a thorough look


jviotti (approved):
Looks good based on what I see, and the whole test suites agree. A small thing to keep in mind (not sure if it still applies) is that Lodash was many times slower than doing the operations directly. We went to great lengths to get rid of Lodash in certain parts of the Contrato library, as the speed improvement was noticeable

jviotti (approved):
I'm happy with how this looks so far!

PR #2817, opened 2019-11-25
core: Add backwards compatible support for versions

jviotti:
Change-type: minor

***

test suite!


ffissore (commented):
Not sure about that SQL statement, the rest looks good to me

ffissore (approved):

PR #2829, opened 2019-11-28
Split webpack bundle in vendor and local

ffissore:
Split webpack bundle in `vendor` and `local`, use HashedModuleIdsPlugin so that the hash reflects each file contents. Vendor bundle is made of our dependencies, so it's unlikely to change: this way browsers can cache it more easily


jviotti (approved):
Looks good code wise, but I might not know the implications in detail

LucianBuzzo (approved):

PR #2844, opened 2019-12-03
Oauth: Add endpoint for balena oauth redirect

karaxuna:
Change-type: minor
Connects-to: #2724

***

test suite!


jviotti (changes requested):

jviotti (changes requested):

jviotti (changes requested):
Its shaping up quite well :+1: I left some small comments here and there. 

I do think that there are parts of this PR that we could already merge separately. For example, you could send a separate PR with the Balena API OAuth base URL declaration (and it's corresponding `Makefile` changes), or even split out the `sync.associate()` function without introducing the new action (as an internal change)

jviotti (changes requested):
I took a much deeper look at it. I left some small comments, but we're getting very close!

PR #2851, opened 2019-12-04
server: Forwards compatible e2e tests with versions support

jviotti:
Change-type: patch

***

test suite!


StefKors (changes requested):
left a comment on the wrong pr

StefKors (approved):
yes, looks good

PR #2853, opened 2019-12-05
ui-components: Backwards compatible version reference support

jviotti:
Change-type: minor

***

test suite!


StefKors (changes requested):
Screenshot 2019-12-05 at 11 46 05

When running a quick test, messages appear twice? one in the `sending` state and once in the posted state

It seems to go away after posting a new message or X amount of time

StefKors (approved):
You are correct, this is also tested in the

PR #2891, opened 2019-12-13
2878 Fix: unread notification count can get cut off

StefKors:
Connects-to: #2878 

***

test suite!


karaxuna (changes requested):

LucianBuzzo (approved):
I'm happy with this, though it would be nice to implement @karaxuna comment about not using fixed width

karaxuna (approved):

PR #2893, opened 2019-12-13
Oauth: Split authorize and associate actions

karaxuna:
Change-type: minor
Connects-to: #2724

***

test suite!


jviotti (approved):
Looks good to me given my latest comments were marked as resolved!

PR #2896, opened 2019-12-16
Add specific index on data->'mirrors' 

ffissore:
Add specific index on `data->'mirrors'` because sysdig reports it's a frequent query and it takes (tens of) seconds


jviotti (approved):
Looks good. If the index creation might take hours, can you ask Mike to leave it running on the Bastion host instead, so I don't risk accidentally stopping it half way through

PR #2927, opened 2019-12-18
default-cards: Set versions in types

jviotti:
Change-type: minor

***

test suite!


StefKors (approved):
looks good to me

PR #2944, opened 2019-12-25
ChatWidget: Fix null is not an object console error

karaxuna:
Change-type: patch
Connects-to: #2805

***

test suite!


LucianBuzzo (changes requested):
Good work! The code looks fine, but can you please amend your commit message to describe the change thats been made: "Fix sentry bug" is far too ambiguous.

LucianBuzzo (approved):

PR #2947, opened 2019-12-31
UI: Account for periods when highlighting usernames

LucianBuzzo:
Fixes #2593

Change-type: patch

***

test suite!


StefKors (dismissed):
This could benefit from having a test which checks if the various usernames are highlighted correctly. (I think similar to the `createPrefixRegExp() match underscore characters`)

grahammcculloch (approved):

PR #2954, opened 2020-01-02
UI: Add click area on message body

StefKors:
Fixes: #2902 
Change-type: patch

This PR increases the clickable area to include the message body for easier navigation.


LucianBuzzo (commented):
This feels a little awkward, for example in a support thread, where it seems that you can click on something but it doesn't do anything. Perhaps we need to use an arrow icon, similar to flowdock, to indicate the click action? Lets talk about this tomorrow, as I think the solution requires a little brainstorming

LucianBuzzo (approved):
We might want to revisit this issue in the future, but I think this is a really good starting point

PR #2958, opened 2020-01-06
UI: update actioncreator stream to handle all target channels

StefKors:
Fixes: #2934
Change-type: patch


LucianBuzzo (approved):

jviotti (approved):
Looks good to me given the test, but it might be worth @LucianBuzzo 's feedback

PR #2972, opened 2020-01-07
Use rabbitmq as message broker

ffissore:
Replace current db based queue implementation with rabbitmq.

The implementation relies a lot on [what was done last time](https://github.com/balena-io/jellyfish/pulls?q=is%3Apr+rabbitmq+is%3Aclosed), with 2 important changes:
- in rabbitmq we don't just push the slug of the action request (that we will use to fetch the actual action request from the db when we'll handle it); instead, we push the whole action request, thus removing the need for the `requests` table and reducing the load on the db
- we don't `poll` the queue to see if a message has been received in the meanwhile; instead, we immediately react when a message is received

Here the details:
- makefile
  - added `start-rabbitmq` task, and conf file to start a local rabbitmq
- scripts:
  - added `scripts/rabbitmq-delete-test-queues.js` that deletes test queues from rabbitmq
  - added `scripts/production/rabbitmq-info.js` that shows info about rabbitmq queues and consumers
- queue:
  - `queue` has been split in `consumer` and `producer`.
  - `consumer` is event based: there no longer is a `dequeue` method that goes to the db. An initialized consumer receives 1 message at a time from rabbitmq and, once processed, it acknowledges rabbitmq.
  - `producer` does not *enqueue* action requests into the db, thus tables `requests` and `locks` are no longer used
 - both `consumer` and `producer` are no longer `EventEmitter` (this is a clean up: there were not emitting events anyway)
- tests:
  - removed all lock and db based queue test code
  - tweaked queue helper so that tests can still call `dequeue`: behind the scenes, the `consumer` is piling up messages in an array, and acknowledging them immediately
  - for each spec file, ava will now also run `./test/max-listeners` which will set process maxListeners to a higher value so that we don't log nodejs complaints (which btw are caused by the logger we use)

Notes: this PR is based on #3032, which has to be merged before this one


mikesimos (commented):
Deploy templates LGTM, however the new secrets values need to be in place before merging this

jviotti (changes requested):

jviotti (approved):
Looks good to me now!

mikesimos (commented):
deploy-templates LGTM

mikesimos (approved):

PR #2974, opened 2020-01-08
SDK: Switch to using unit tests with nock

LucianBuzzo:
Most e2e tests with the SDK duplicate behaviour already covered in the
UI e2e tests. This commit strips down the SDK tests, and runs them as
unit tests with a mock API using nock.

***

test suite!


StefKors (approved):
Looks good! excited for those speed gains

PR #2977, opened 2020-01-08
deploy-templates: Deploy livechat

jviotti:
Change-type: minor

***

test suite!


richbayliss (approved):
I dont see anything obviously wrong K8s-wise, and Katapult will just invoke those changes on the cluster anyway, so I see no reason to approve it.

karaxuna (approved):

PR #2979, opened 2020-01-08
Make it easy to run servers only (postgres, redis) using docker-compose

ffissore:
Instead of using custom and local docker commands, or use a locally installed redis or postgres, we should use the exact same thing we use in docker-compose

We can now use `make compose-servers`

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -8,8 +8,6 @@
 	start-server \
 	start-worker \
 	start-tick \
-	start-redis \
-	start-postgres \
 	test-unit \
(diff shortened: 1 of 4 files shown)


jviotti (changes requested):
The reason I'm running the services natively on macOS is due to the extremely poor performance of Docker Compose for Mac. Can we try to improve the BoB situation first? Running everything on Docker Compose will be super slow for me and will drain my battery life when on the go.

PR #2980, opened 2020-01-08
Rabbitmq on production (just the server)

ffissore:
On production, start a new, unused rabbitmq instance, so we can check its wiring

Connects-to: #2796
Change-type: patch


jviotti (approved):

mikesimos (approved):
Looks good :+1:

PR #2982, opened 2020-01-09
Fix docker-compose API endpoint vars

joshbwlng:
Change-type: patch

***

Fixes `docker-compose` access issues noted in issue https://github.com/balena-io/jellyfish/issues/2973

***

test suite!

### DIFF
--- a/README.md
+++ b/README.md
@@ -35,6 +35,7 @@ Compose](https://docs.docker.com/compose/) installed. First build all the
 containers:
 
 ```sh
+npm install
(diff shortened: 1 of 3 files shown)


jviotti (approved):
Looks good, but you have to update the `docker-compose.tpl.yml` and then `make docker-compose.yml`

LucianBuzzo (commented):
Looks like you just need to add a `Change-type` footer and sign your commit and the versionist check should pass

PR #3006, opened 2020-01-14
Run 1 replica of jellyfish-ui, instead of 3

ffissore:
Run 1 replica of jellyfish-ui, instead of 3. Jellyfish-ui is an nginx container serving static files, so running multiple replicas provides no advantage

Change-type: patch


jviotti (approved):

LucianBuzzo (approved):

joshbwlng (approved):

mikesimos (changes requested):
We need at least 2 replicas for high availability. Running 1 replica means running a risk of service disruption/downtime, every-time an underlying node fails, or on potential pod failure. We should close/not merge this PR.

PR #3020, opened 2020-01-16
Accept balena API oauth request

karaxuna:
Change-type: patch

***

test suite!


jviotti (approved):
I left a small comment about how we can improve the assertion rather than dropping it, and it all looks fine apart from that

PR #3027, opened 2020-01-17
action-server/server: Fix production scripts copying

jviotti:
Change-type: patch

***

test suite!


ffissore (approved):

LucianBuzzo (commented):
@ffissore @jviotti do we really need these scripts? It seems like a big hack :/

PR #3032, opened 2020-01-20
Rabbitmq: updated deploy templates

ffissore:
- Move k8s namespace creation/declaration at the top of the bundle generated by katapult, as it seems to be wrong doing otherwise
- Don't give ports a name: we don't use them and it's error prone
- Fix keyframe structure for rabbitmq
- Add script to query the status of rabbitmq

Change-type: patch

***

test suite!


jviotti (approved):
Looks good to me given my k8s knowledge :)

joshbwlng (approved):

mikesimos (approved):
lgtm

PR #3047, opened 2020-01-24
UI: Use immutability-helper in Redux reducers

grahammcculloch:
Ref: #3029 

This PR refactors the way the UI app state is updated in the Redux reducers. It makes use of  [immutability-helper](https://github.com/kolodny/immutability-helper) to ensure application state is immutable (within the reducers) but is still updated fast. 

Some more unit tests have been added to verify/explain the behavior of some of the more complex reducers. _Note: `lib/chat-widget` is now also included in the paths to run through babel for the unit test pre-processing. Before there were no unit tests in this directory._

Changelog-Entry: Performance improvements under the hood
Change-type: patch

***

**Please note: this is a significant (but simple) refactor of application-critical code by someone who is new to the code-base (me!). I'm working on the assumption that e2e tests (in addition to the unit tests I've added) are sufficient to verify the absence of regression bugs.**

### DIFF
--- a/apps/ui/core/store/reducer.js
+++ b/apps/ui/core/store/reducer.js
@@ -5,67 +5,11 @@
  */
 
 import clone from 'deep-copy'
+import update from 'immutability-helper'
 import * as _ from 'lodash'
 import * as redux from 'redux'
 import uuid from 'uuid/v4'
 import actions from './actions'
-
-const viewsReducer = (state, action) => {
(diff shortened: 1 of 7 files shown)


LucianBuzzo (changes requested):
One small change needed I think, otherwise looks great!

LucianBuzzo (approved):

PR #3050, opened 2020-01-28
Fix link-constraints to use correct verbs

StefKors:
Change-type: patch


LucianBuzzo (changes requested):
Small change needed, but otherwise looks good :+1:

jviotti (approved):

LucianBuzzo (approved):

PR #3052, opened 2020-01-29
Add info about fixing webpack memory error to README

ljewalsh:
***

test suite!


LucianBuzzo (changes requested):
Thanks for adding this! I can see that the `versionist` check is failing, this is because we need to format commits with some specific footers for our CI process to function correctly. Take a look through https://github.com/balena-io/process/blob/master/process/Commit_and_PR_Guidelines.md for details :+1:

PR #3072, opened 2020-02-05
add mailgun integration

ljewalsh:
With sendEmail function so we can send emails to users

### DIFF
--- a/ARCHITECTURE.md
+++ b/ARCHITECTURE.md
@@ -137,6 +137,10 @@ module provides a set of functions that the backend uses for logging purposes.
 - The logger is able to pipe logs to a central location when running in
 	production
 
+### [`lib/mail`](https://github.com/balena-io/jellyfish/tree/master/lib/mail)
+
+The mail library consists of a mailgun integration which can be used to send emails to jellyfish users
+
 ### [`lib/queue`](https://github.com/balena-io/jellyfish/tree/master/lib/queue)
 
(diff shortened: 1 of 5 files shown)


grahammcculloch (commented):
Just some little initial comments!

joshbwlng (approved):

ffissore (approved):

jviotti (changes requested):
The code in `lib/sync/integrations` is about syncing bi-directional data and the classes in there have a predefined interface that we are not respecting here (i.e. `.translate()` and `.mirror()`) which operates on cards.

Can we refactor the implementation so that:

- The `.sendEmail()` function becomes `.mirror()`, matching the arguments of the `.mirror()` methods in the other integrations
- We have a new `email` card type that represents an email being sent
- Make `.mirror()` operate on `email` cards. Basically it gets an `email` card as an argument and sends an e-mail in response to that card
- Add a mailgun mirror action like we do for other integrations
- Add a triggered action so that we call such function when we insert `email` cards into the system

?

LucianBuzzo (commented):
The code looks good, but I think it needs to live in a different file location. This code doesn't "sync" anything. I think it might be sensible to put it under `lib/mail`. Also, is there any reason why we aren't using a pre-existing package? e.g. https://www.npmjs.com/package/mailgun-js

LucianBuzzo (approved):

jviotti (approved):

PR #3082, opened 2020-02-07
Use graphile-worker instead of our own implementation of a db based queue

ffissore:
Replace current db based queue implementation with graphile-worker.

2 important characteristics of this implementation:
- we don't just push the slug of the action request (that we will use to fetch the actual action request from the db when we'll handle it); instead, we push the whole action request, thus we don't need the `requests` table any more
- we don't `poll` the queue to see if a message has been received in the meanwhile; instead, we immediately react when a message is received

Here the details:
- queue:
  - `queue` has been split in `consumer` and `producer`.
  - `consumer` is event based: there no longer is a `dequeue` method that goes to the db. An initialized consumer receives 1 message at a time.
  - graphile-worker has its own db and tables, so we don't need tables `requests` and `locks` any more
 - both `consumer` and `producer` are no longer `EventEmitter` (this is a clean up: they were not emitting events anyway)
- tests:
  - removed all lock and db based queue test code
  - tweaked queue helper so that tests can still call `dequeue`: behind the scenes, the `consumer` is piling up messages in an array


jviotti (approved):
A lot of this changes are due to changing the queue to a consumer/producer interface, which I reviewed on the RabbitMQ PR. Looks good to me!

nazrhom (approved):

PR #3089, opened 2020-02-11
Add role-user-operator to default cards when server is run

ljewalsh:
We need an operator role in order for certain admin-level users to have additional permissions to that of the community user. The expectation is that this role would sit alongside the community role on a user (i.e. they would have both the community and operator role)

At the moment, the only difference between the operator and the community role is that the operator can view the view-all-user page and thus see and create users

I added some tests to ensure that this is the case. Also moved community-role-related security tests into their own file to make things a little cleaner


grahammcculloch (commented):
Presumably users can be assigned multiple roles? If so, the operator role should _only_ define the permissions specific to the operator role. All the community role permissions should not be included. If a user is granted the operator role it would not be _instead_ of the community role but in _addition_ to it. Or have I completely misunderstood how Jellyfish roles work?

LucianBuzzo (approved):

Issue #3054, opened 2020-01-29
Serve livechat bundles using a CDN


LucianBuzzo:
In some geographical locations, loading the UI can take a really long time due to high latency and download times. Using a CDN should be a quick win that can help improve this situation

PR #3093, opened 2020-02-11
Improve the way webpack packs bundles

ffissore:
Improve the way webpack packs bundles, in order to make them more cache friendly

With #2829 we splitted bundles in main and vendor. ~However according to [this article](https://developers.google.com/web/fundamentals/performance/webpack/use-long-term-caching) it was missing a critical piece called `runtimeChunk`~
~Article found via [hackers news tweet](https://www.debugbear.com/blog/performant-front-end-architecture)~

After some additional research, I've found [this webpack doc](https://webpack.js.org/guides/caching/) about caching: while google is the usual know-it-all, it made more sense to follow webpack instructions instead

This PR relates to #3054

### DIFF
--- a/apps/livechat/webpack.config.js
+++ b/apps/livechat/webpack.config.js
@@ -7,9 +7,9 @@
 /* eslint-env node */
 
 const HtmlWebpackPlugin = require('html-webpack-plugin')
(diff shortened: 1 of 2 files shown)


StefKors (approved):
LGTM

LucianBuzzo (approved):

PR #3096, opened 2020-02-12
Add faq card type to default cards

ljewalsh:
The new card has three fields:
- name (for the FAQ question)
- answer (markdown so we can add complex styling to our answers)
- topic (so we can group FAQ questions and answers together)

For now topic is a simple string. I was not sure whether the grouping could occur yet so I left everything as a simple view

There also didn't seem to be any tests for the other cards. I'd be happy to add some simple tests (e.g. did this card load when the server starts?). Let me know if this is something you'd like

test suite!


LucianBuzzo (changes requested):

grahammcculloch (changes requested):
@ljewalsh I think you'll need/want to squash the 3 commits into one (to appease the VersionBot Gods). I just found [this excellent guide](https://www.internalpointers.com/post/squash-commits-into-one-git) that walks you through doing it in case (like me) you aren't super familiar with the process.

ffissore (approved):

PR #3098, opened 2020-02-12
View templates

ffissore:
Introduce view templates, a way to query by views which are json-e templates, saving callers from building complex query objects every time and instead use a view template referenced by slug and provide its params

Available via `sdk.view`, for example 
```js
sdk.view('view-all-by-type@1.0.0', {
  types: [ 'view', 'view@1.0.0' ]
})
```

Connects-to: #3097
Change-type: patch


jviotti (commented):
Can we add an e2e test that makes use of a parametric view, just to make sure it all ties together correctly?

jviotti (approved):

PR #3102, opened 2020-02-13
test: Move sync translate and integration tests to balenaCI

LucianBuzzo:
Change-type: patch

***

test suite!

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -86,7 +86,7 @@ jobs:
           paths:
             - /images.tar
 
-  unit_and_sync_tests: &unit_and_sync_tests
+  sync_tests: &sync_tests
     <<: *job_defaults
 
     steps:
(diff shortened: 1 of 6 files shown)


joshbwlng (approved):

joshbwlng (changes requested):
So have you talked with @nazrhom about this? I talked with him last night and what the balenaCI team has decided to do is run all tests that do no require secrets (unit, etc.) on balenaCI through a `docker-compose.test.yml` file and keep all tests that talk to external services on CircleCI. This obviously conflicts with what you're trying to do here. The reason to keep external service tests on CircleCI is because balenaCI doesn't support project-level secrets, but I guess if you think it's OK to just put all of the tokens/keys in source files (even if they are fakes), then we could do the same for the `docker-compose.test.yml` setup?

PR #3108, opened 2020-02-14
Add blank space to README to redeploy

karaxuna:
Change-type: minor

***

test suite!


nazrhom (changes requested):
@karaxuna you can create an empty commit with `--allow-empty`, also an empty commit is probably a patch change rather than a minor :P

PR #3111, opened 2020-02-17
Update open-balena-base and disable circleci docker_layer_caching

ffissore:
* Update open-balena-base to latest 9.2.0
* Disable circleci docker_layer_caching when not needed, as it costs 200 credit per job run https://circleci.com/docs/2.0/docker-layer-caching/

For the record, the only job that needs `docker_layer_caching` is `compose-build`. The other ones use the saved docker layers that we restore each time, so no need to spend those 200 credits, as we don't leverage `docker_layer_caching` in those cases

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -188,8 +188,7 @@ jobs:
       - attach_workspace:
           at: .
 
-      - setup_remote_docker:
-          docker_layer_caching: true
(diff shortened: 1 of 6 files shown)


joshbwlng (approved):
LGTM

grahammcculloch (approved):

PR #3114, opened 2020-02-17
Small improvements

ffissore:
A couple of small improvements, not related to any specific on going activity, worth pushing to master

* Postpone `cloneDeep` in streams schema validation until we actually need it
* Wait a bit before tearing down queue tests, as kernel is still doing stuff in background. This fix might put an end to those SIGILL failures we have seen any now and then


karaxuna (approved):
LGTM :+1:

PR #3131, opened 2020-02-18
Use caching headers for livechat assets

ffissore:
Since #3093, livechat assets are split as well, so we can now use caching headers such as 'expires' and 'cache-control'


karaxuna (approved):
LGTM :+1:

Issue #3134, opened 2020-02-19
webpack DefinePlugin causes vendor bundle file name to change


ffissore:
When building UI for different JF releases, the vendor bundle file name (and contents) always change, even if no changes were made to the dependencies

A `diff` between two `vendors.*.js` files shows that the only difference is the version of the UI.

This change of filename prevents browsers from caching the file, which is ~8Mb uncompressed.

Having failed to find a solution, I've asked stackoverflow: https://stackoverflow.com/q/60298365/635621

PR #3137, opened 2020-02-19
Don't use `process.env` as UI env vars container

ffissore:
Don't use `process.env` as UI env vars container, because it interferes with some third party modules using it: the result is a different `vendors` bundle for each release of the UI, even if no
changes involved third party libraries.

If you cherry-pick this changeset, and apply it to two different releases, v24.70.2 and v24.70.1, the output of `make build-ui` will be the same `vendors` and `runtime` bundles, and two different `main` bundles (the only difference being jellyfish version).

If instead you run `make build-ui` on the two mentioned versions, without this changeset, the output will be two different `main` **and** `vendors` bundles. See [gist](https://gist.github.com/ffissore/c242669a37254a3eee2cae0e94045648).

The reason is that webpack will replace every occurrence of `process.env` with the object defined with `DefinePlugin`, **even for third party libraries**: list includes debug, bluebird, monaco-editor, react-chartjs-2, react-trello, styled-components.
Take a look at [this line from "debug" module](https://gist.github.com/ffissore/c242669a37254a3eee2cae0e94045648#file-vendors-diff-L8) for example: the [original code](https://github.com/visionmedia/debug/blob/master/src/browser.js#L217) is `process.env.DEBUG`, and webpack replaces it with `Object({...our keys...}).DEBUG`.

Connects-to: #3134
Change-type: patch

### DIFF
--- a/apps/livechat/webpack.config.js
+++ b/apps/livechat/webpack.config.js
@@ -60,7 +60,7 @@ const config = mergeConfig(baseConfig, {
 
 		new DefinePlugin({
 			/* eslint-disable no-process-env */
-			'process.env': {
+			env: {
 				API_URL: JSON.stringify(process.env.API_URL),
 				NODE_ENV: JSON.stringify(process.env.NODE_ENV)
 			}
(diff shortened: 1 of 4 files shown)


StefKors (approved):
LGTM

LucianBuzzo (approved):

PR #3140, opened 2020-02-20
Lazy logging error messages

ffissore:
On your machine, save the following snippet in JF repo root as `test.js`, then run `node test.js`
```js
const assert = require('./lib/assert')

const obj = {
	simple: 'object'
}
const name = 'test'

const assertObj = (obj) => {
	assert.INTERNAL(null, true, Error, `Missing context: ${JSON.stringify(obj)}`)
}
const assertString = (str) => {
	assert.INTERNAL(null, true, Error, `Missing context: ${str}`)
}

const TEN_MILLION = 10000000

let now = new Date()
for (let i = 0; i < TEN_MILLION; i++) {
	assertObj(obj)
}
console.log('object', new Date() - now)
now = new Date()
for (let i = 0; i < TEN_MILLION; i++) {
	assertString(name)
}
console.log('string', new Date() - now)
```

With the current version of `assert`, it will print something like this
```
object 2369
string 40
```
That's 2.3 seconds for **not** throwing errors that involve stringifying objects. The more complex `obj` is, the longer it will take. That's all wasted CPU, used to execute `JSON.stringify(obj)`

Now change `test.js` like so

```diff
@@ -6,10 +6,14 @@ const obj = {
 const name = 'test'
 
 const assertObj = (obj) => {
-       assert.INTERNAL(null, true, Error, `Missing context: ${JSON.stringify(obj)}`)
+       assert.INTERNAL(null, true, Error, () => {
+               return `Missing context: ${JSON.stringify(obj)}`
+       })
 }
 const assertString = (str) => {
-       assert.INTERNAL(null, true, Error, `Missing context: ${str}`)
+       assert.INTERNAL(null, true, Error, () => {
+               return `Missing context: ${str}`
+       })
 }
 
 const TEN_MILLION = 10000000
```

and run it once more: it will print something like this
```
object 13
string 14
```
Much faster, ~180 times, when the error message contains `JSON.stringify(obj)`, because it's not executed. Faster but not so much when it contains a simple string.

Thus, I changed all calls to `assert.USER` and `assert.INTERNAL` to use lazy functions when the error message is not a simple string template.

Additionally, some improvements to `logger` module:
* sharing `transport` instances between `Logger`s
* exit immediately if log level doesn't match

### DIFF
--- a/lib/assert/index.js
+++ b/lib/assert/index.js
@@ -4,7 +4,11 @@
  * Proprietary and confidential.
  */
 
-const createError = (context, error, message, options = {}) => {
+const _ = require('lodash')
+
+const createError = (context, error, messageOrFunction, options = {}) => {
+	const message = _.isFunction(messageOrFunction) ? messageOrFunction() : messageOrFunction
+
 	// eslint-disable-next-line new-cap
 	const instance = new error(message)
 	instance.context = context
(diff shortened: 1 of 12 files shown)


jviotti (approved):
Good find!

PR #3143, opened 2020-02-21
WIP: cleanup test helpers

ljewalsh:
***

test suite!


ffissore (changes requested):
This PR includes two very different commits. One is a refactoring of test helpers, the other is (I think) a fix for #3144. We want both, but it would be better to have them in two different PRs: main reason is that it makes it easier to review it, secondary is that, if discourse tests fail, we want to be sure it's not because of the helpers refactoring
Also, if the discourse commit is for #3144, it should mention it in the commit message

LucianBuzzo (changes requested):

PR #3152, opened 2020-02-24
WIP: ui: Add support for ownership of support threads

grahammcculloch:
You can now easily set the owner of a support thread via a dropdown button on the Card Actions area.

Change-type: minor

***

CardOwnership Progress 2

## Card Ownership changes

This PR replaces the concept of 'Primary owner' and 'Secondary owners' with 'Owner' and 'Backup owners'. This is a breaking change in so far as any existing 'primary' and 'secondary' owners will no longer be shown in the UI. But the understanding is that these links (used on just Projects, Accounts and Contacts) were not being used yet anyway.
* The 'Primary owner' tab has been removed from Account, Project and Contact card lenses. 
* The 'Secondary owners' tab has been renamed to 'Backup owners'
* The following cards support card ownership and therefore have the new 'Card Owner' dropdown control in the 'Card Actions' section:
  * `account`
  * `contact`
  * `project`
  * `support-thread` 

## Link changes
As part of this work, I've had to update the way link card slugs are generated:
* they now contain a uuid to ensure they won't conflict with a previously created _inactive_ link card with the same verb, fromId and toId.
* when linking cards, we first query for existing links with that verb, fromId and toId. If found, we just return the existing card. Otherwise we create a new link card - which will have a unique slug thanks to the uuid part of it.

## SetupProvider
This PR also fixes the `SetupProvider` component. The react-redux `Provider` component is now above the `SetupProvider` component in the component hierarchy and so all the redux actions (generated from the action creators) can be passed down in the React context (along with `sdk` and `analytics`).

This change simplifies various down-stream components as they now have access to all the Redux actions and we don't need to manually pass them through or bind them specifically each time. There's probably various other components where this could be tidied up a lot too!

### DIFF
--- a/apps/server/default-cards/contrib/account.json
+++ b/apps/server/default-cards/contrib/account.json
@@ -92,13 +92,8 @@
           "type": "contact"
         },
         {
-          "title": "Primary owner",
-          "link": "has primary owner",
-          "type": "user"
-        },
-        {
-          "title": "Secondary owners",
-          "link": "has secondary owner",
+          "title": "Backup owners",
+          "link": "has backup owner",
           "type": "user"
         }
       ]
--- a/apps/server/default-cards/contrib/contact.json
+++ b/apps/server/default-cards/contrib/contact.json
@@ -96,13 +96,8 @@
           "type": "account"
         },
         {
-          "title": "Primary owner",
-          "link": "has primary owner",
-          "type": "user"
-        },
-        {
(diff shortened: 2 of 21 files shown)


StefKors (changes requested):
Looks great, especially the helpers for the tests are super useful. Just the added test case requires some updates I think.

StefKors (changes requested):

PR #3175, opened 2020-02-26
front: Don't consider private inboxes

jviotti:
If a conversation starts on a private personal inbox and is then moved
to i.e. support, then the conversation will belong to both inboxes at
the same time.

When looking what inbox a thread belongs to, we shouldn't take into
account private ones, as we don't know what to do with them.

Change-type: patch

***

test suite!


ljewalsh (changes requested):
Looks good to me. I wonder if we can put this in a named variable to make it clear what it's doing for future devs.

Also, should we have a test for this?

LucianBuzzo (approved):

ffissore (changes requested):
Do we have a test for this case already? If not, can we add one?

ffissore (approved):

PR #3185, opened 2020-02-27
Add test data cleanup scripts and CircleCI job

joshbwlng:
Change-type: patch

***

Add old data clean up scripts:
- Close GitHub issues created on the [test repo](https://github.com/product-os/jellyfish-test-github/issues) that are two or more days older
- Delete Front conversations created in the test inboxes that are two or more days older

This PR will also add two make commands to perform the above and automatically run these scripts in a new CircleCI job.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -781,6 +781,18 @@ jobs:
           name: Test
           command: cd ./scripts/eslint-plugin-jellyfish && npm install && npm test
 
+  clean_old_test_data:
+    <<: *job_defaults
+    steps:
+      - checkout
+      - attach_workspace:
+          at: .
+      - run:
+          name: Clean up old test data
+          command: |
+            make clean-github
+            make clean-front
+
 workflows:
     version: 2
     build:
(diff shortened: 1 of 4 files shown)


grahammcculloch (changes requested):
@joshbwlng great work! I really like the judicious use of comments - makes it really clear and easy to follow!

I've just focused on the node.js code and added some nit-pick comments mainly about slightly neater/clearer ways to write the same code (as you mentioned this was your first node.js PR). 

Main thing to potentially change is the UTC issue.

grahammcculloch (changes requested):
Updates all look good!

One final request: I think it would be worth adding fuller comments at the top of the `github-close-issues.js` and `front-delete-conversations.js` files that detail:
* An example of how to call the script (even if you're explicitly saying that no parameters are required)
* All environment variables used by the script and what they mean
* A short summary of what the script does (including in this case the fact that only conversations/issues older than x days are deleted)

grahammcculloch (approved):
Changed to 'approve' to make the documentation comments optional!

PR #3186, opened 2020-02-27
Allow env to be undefined

StefKors:
Because react isn't compiled by Webpack, env vars aren't included during unit tests. This PR allows these to be undefined if "env" doesn't exist.

This error showed up after updating AVA and Sinon in PR #3160

Change-type: patch

### DIFF
--- a/apps/ui/environment.js
+++ b/apps/ui/environment.js
@@ -15,18 +15,18 @@ export const isProduction = () => {
 }
 
 export const sentry = {
-	dsn: env.SENTRY_DSN_UI || '0'
+	dsn: (typeof env === 'undefined') ? '0' : (env.SENTRY_DSN_UI || '0')
 }
 
 export const api = {
-	prefix: env.API_PREFIX || 'api/v2',
-	url: env.API_URL || window.location.origin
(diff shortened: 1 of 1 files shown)


grahammcculloch (commented):
Whilst this approach may prevent unit tests from falling over, there are various ways to allow tests to make use of environment variables. See [this thread](https://github.com/vuejs/vue-test-utils/issues/193) for some examples.

I think in the long run we'll be better off if we _can_ support environment variables in UI unit tests.

ffissore (approved):
Yeah, double checked even using `_.get` and `pirates` keeps complaining. What is pirates?
I'm good to have this code merged

grahammcculloch (changes requested):
Just needs a small logic tweak to be fully equivalent to the previous logic.

grahammcculloch (commented):
Just spotted one issue that might have slipped through the refactor (see inline comment)

grahammcculloch (changes requested):
Er... still not quite right! (Again, see inline comment)

grahammcculloch (approved):

PR #3187, opened 2020-02-27
balena-api: Check that geoip result is defined

jviotti:
This can be the case if the ip doesn't resolve to any valid location.

Change-type: patch
Fixes: https://sentry.io/share/issue/266bb268586a40e092c37777872ddca5/

***

test suite!


grahammcculloch (approved):

joshbwlng (approved):

ffissore (changes requested):
Do we have a test for this case already? If not, can we add it?

ffissore (approved):

PR #3188, opened 2020-02-27
balena-api: Retry .whoami() handler on 429

jviotti:
Change-type: patch
Fixes: https://sentry.io/share/issue/ffceb80789074ce3bce0b7f7f50d952f/

***

test suite!


grahammcculloch (approved):

joshbwlng (approved):

ffissore (changes requested):
Do we have a test for this case already? If not, can we add one?

ffissore (approved):

PR #3189, opened 2020-02-27
github: Fallback for issue mirror URL

jviotti:
Change-type: patch
Fixes: https://sentry.io/share/issue/b79ad9abb1c74de5b3d8de815ed07e3a/

***

test suite!


grahammcculloch (commented):
Just one question for you...

ffissore (changes requested):
Do we have a test for this case already? If not, can we add one? (sorry for repeating this ad nauseam, it must be the friday mood)

grahammcculloch (approved):

ffissore (approved):

PR #3202, opened 2020-03-01
Add faq card type to default cards

ljewalsh:
The new card has three fields:

- name (for the FAQ question)
- answer (markdown so we can add complex styling to our answers)
- topic (so we can group FAQ questions and answers together)

For now topic is a simple string. I was not sure whether the grouping could occur yet so I left everything as a simple view


grahammcculloch (changes requested):
Could you add an e2e test that adds a FAQ card via the SDK and then verifies it is displayed when you navigate to this view?

grahammcculloch (approved):
We've decided to merge this PR now as the CI tests have finally all passed. We can write a e2e test in a new PR. The new FAQ functionality has been manually tested.

PR #3210, opened 2020-03-02
Introduce AWS managed redis cluster

ffissore:
Jellyfish at the moment uses a single redis instance, that we deploy via k8s. For better reliability, we should switch to an AWS managed redis cluster


cmfcruz (commented):
I think we should also add the following property to ReplicationGroup to allow automatic failover.  The default value is false:

```
AutomaticFailoverEnabled: true
```

cmfcruz (approved):

ab77 (approved):

PR #3212, opened 2020-03-02
circleci: Skip sync integration tests on UI-only changes

jviotti:
Change-type: patch

***

test suite!

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -117,28 +117,28 @@ jobs:
 
       - run:
           name: Sync Tests Outreach Translate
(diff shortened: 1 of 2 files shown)


joshbwlng (changes requested):
Nice, this was kind of what I was going for with https://github.com/product-os/jellyfish/issues/3195
Being able to skip sync tests when unnecessary should help unblock us quite a bit 👍 

I'm assuming you're covering this by choosing where you call it in the CircleCI config, but would it also make sense to add logic to skip when docs or scripts under `scripts/` etc. are updated? Probably not needed, but might be nice for completion sake.

Also, it would make it easier to follow what the bash script is doing step by step if there were a few comments.

LucianBuzzo (approved):

joshbwlng (approved):

PR #3213, opened 2020-03-03
dev: Check node version after npm install

grahammcculloch:
Added a simple script that will error if the node version is less than the one specified in `engines` in `package.json`. This script is called by the postinstall routine.

Change-type: minor

***

We now specify node version requirements in `package.json`:
```json
"engineStrict": true,
"engines": {
  "node": ">=10.14.2"
},
``` 

If you run `npm install` (or `npm ci`) now and your node version is too low. you'll see something like the following:

check_node_version


ljewalsh (approved):
LGTM

joshbwlng (changes requested):

joshbwlng (approved):

PR #3220, opened 2020-03-03
ui: Don't setup jsdom in e2e tests

jviotti:
There should be no reason to make the e2e suite definition files browser
contexts, as we are already driving a Puppeteer instance that is a real
browser.

Change-type: patch

***

test suite!


grahammcculloch (changes requested):
Ironically @StefKors is just working on a PR that will make use of jsdom (via `browser-env`) for **Unit Tests**. Any chance you could leave in the package dependency on `browser-env`? He'll probably create his own `test/unit/browser-setup.js`.

grahammcculloch (approved):

joshbwlng (approved):

PR #3228, opened 2020-03-03
test: Execute after/before hooks serially

jviotti:
According to the docs, before/after hooks run all in parallel unless
they are explicitly prefixed by "serial".

Making them serial will help us debug cases like
https://circleci.com/gh/product-os/jellyfish/74925, where before/after
hooks blow up and we never get a stack trace as we don't enable
`--fail-fast` in the CI.

Change-type: patch

***

test suite!

### DIFF
--- a/lib/chat-widget/store/reducer.spec.js
+++ b/lib/chat-widget/store/reducer.spec.js
@@ -10,7 +10,7 @@ const {
 	createReducer
 } = require('./reducer')
 
-ava.beforeEach((test) => {
+ava.serial.beforeEach((test) => {
 	test.context.initialState = {
 		product: 'jelly-chat-test',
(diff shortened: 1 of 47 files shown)


ljewalsh (commented):
Can you explain this error Juan? It seems to me that we are creating unique databases for each test so this shouldn't occur/ the serialisation shouldn't be necessary. Is this because we connect to the default db?

joshbwlng (commented):
Do we need all of these to run serially?

grahammcculloch (commented):
One more question: if we make all `beforeEach` calls run serially, do we prevent the tests themselves from running in parallel? Or can you still mix serial `beforeEach` calls with parallel test methods?

ffissore (approved):

PR #3229, opened 2020-03-03
worker/action-server: Store trigger slugs

jviotti:
The main motivation is to attempt to provide better error messages when
triggered actions fail.

Change-type: patch

***

test suite!


ljewalsh (approved):
LGTM

PR #3231, opened 2020-03-04
Extract worker tests into their own test files

ljewalsh:
Just moving tests from one large test file into smaller test files for each action. Makes it easier to maintain and think about

- I went through an grep'd each test description to make sure I hadn't lost any in the process


grahammcculloch (approved):
Nice! I'm a big fan of breaking things down into smaller chunks. And 5000 lines is too long for one file IMHO!

PR #3232, opened 2020-03-04
Verify parameters on POST signup

joshbwlng:
Change-type: patch

***

ref: https://github.com/product-os/jellyfish/issues/3159

Add some simple string checks on `POST /api/v2/signup` request parameters.

### DIFF
--- a/apps/server/http/routes.js
+++ b/apps/server/http/routes.js
@@ -626,6 +626,21 @@ module.exports = (application, jellyfish, worker, producer, options) => {
 			password
 		} = request.body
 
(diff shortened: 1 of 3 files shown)


ffissore (changes requested):
Can we add a test case for this check? Or do we already have one?

grahammcculloch (approved):
LGTM - great work @ffissore for requesting a test and @joshbwlng for writing some lovely simple tests!

ffissore (changes requested):

ffissore (approved):

PR #3251, opened 2020-03-04
test: Don't run translate tests without tokens

LucianBuzzo:
Testing sync translate code without tokens is an unnecessary step.

Change-type: patch

***

test suite!


jviotti (approved):
Looks good, but some of the sync tests seem to be failing with your change

PR #3259, opened 2020-03-05
ui: Refactor SetupProvider higher-order-component

grahammcculloch:
This change makes all redux actions available via the withSetup HOC and the useSetup hook.

Change-type: patch

***

The `actions` prop defined in `SetupProvider` was never actually populated - so it could not be used by components using `withSetup` or `useSetup`. This PR remedies this. Note that the React-Redux `Provider` component is now upstream of the `SetupProvider` component to allow the latter to make use of the Redux store.

### DIFF
--- a/apps/ui/index.jsx
+++ b/apps/ui/index.jsx
@@ -22,7 +22,8 @@ import {
 import {
 	analytics,
 	sdk,
-	store
+	store,
+	actionCreators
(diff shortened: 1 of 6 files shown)


karaxuna (commented):
Chat widget also uses SetupProvider. Can you please check if chat does not break because of this change?

PR #3262, opened 2020-03-05
Get rid of code coverage support

jviotti:
We are losing this already by moving (part of) the test suite to Balena
CI (https://github.com/product-os/jellyfish/pull/3253), and not doing it
anymore will likely make the tests run faster.

This commit can always be reverted if we want to figure out how to get
this back on Balena CI.

Change-type: patch

***

test suite!


ffissore (approved):
We never used coverage as a blocker, or a warning so removing it is good for me

PR #3264, opened 2020-03-05
sync: Add bigger test timeouts in mirror tests

jviotti:
Looks like some test cases may take a really long time due to rate
limiting, but they will still pass given enough time.

Change-type: patch

***

test suite!


karaxuna (approved):

StefKors (changes requested):
Is it possible to use `beforeEach` for this?
https://github.com/avajs/ava/blob/master/docs/01-writing-tests.md#before--after-hooks

StefKors (approved):
👍

PR #3266, opened 2020-03-05
circleci: Improve overall test speed

jviotti:
- Properly enable Docker layer caching on jobs dealing with Docker
Compose (confirmed that it works this time around) and use it instead of
our custom Docker layer cache logic
- Increase the resource class of the jobs that will do Docker builds
- Don't bring in `node_modules` from the `build` job into the jobs that
don't need it
- Build Docker Compose on every e2e test job *while* the `build` job
still runs, so we get better parallelism instead of having the most
expensive jobs also all blocked by a single expensive build job.
Reducing the amount of data we move across jobs also makes a decent dent
in the build time

Change-type: patch

***

test suite!

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -61,31 +61,6 @@ jobs:
           paths:
               - node_modules
 
-  compose-build:
-    <<: *job_defaults
-
-    resource_class: xlarge
-
-    steps:
-      - checkout
-
-      - setup_remote_docker:
-          docker_layer_caching: true
-
-      - run:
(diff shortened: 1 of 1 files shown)


ffissore (commented):
I have a doubt. I see you don't run `compose-build` any more, and instead rely on `compose-up` only. So docker will reuse images built from a different workflow of a different PR, and use those as "images under test". Is that correct?

LucianBuzzo (approved):

PR #3267, opened 2020-03-06
Refactors action handlers into individual files

ljewalsh:
No code changes, just moving into their own files


grahammcculloch (changes requested):
Just one small suggested change.

Also, I wonder if it would be cleaner/more consistent if you used the pattern
```javascript
module.exports = {
  handler
}
```
in each action handler file rather than some of them using
```javascript
module.exports = handler
```

It just means the person adding a reference to an action handler doesn't need to look into the file to see how it is exported - they know what the pattern is and they know its consistent.

grahammcculloch (approved):
Nice :-)

PR #3279, opened 2020-03-09
add request-password-reset, as well as complete-password actions and handler

ljewalsh:
Adds the action-request-password AND action-complete-password-reset actions and handlers to action-library. 

Request-password-reset meets following criteria:

- Confirms that a user with the matching email is found
- Confirm that the user already has a password hash
- Confirms that the user is active (not disabled)
- If all are confirmed, generate a password-reset code and create a password-reset card (with the code attached)
- Also create the reset-password email body and fire off an action to send the email
- If the action fails, we respond positively regardless (200 response)

To do:

- [x]  ensure resetTokens are unique between users with the same password
- [x]  wait for `RESET_PASSWORD_SECRET_TOKEN` env variable to be set on Jellyfish prod


grahammcculloch (changes requested):
Nice job, this is quite involved! Loving the detailed tests 👍 

Sorry - I left a bunch of comments. Mostly really small things though - should be easy to tweak.

LucianBuzzo (changes requested):

LucianBuzzo (approved):

grahammcculloch (approved):

PR #3287, opened 2020-03-09
Add action-send-email to action-library

ljewalsh:
The action includes toAddress, fromAddress, body, and subject arguments

It uses the mailgun integration to send an email to the toAddress


grahammcculloch (commented):
Looking good!

Couple questions:
1. Should we support sending email to multiple recipients (ie. `toAddress` can be an array)? Or is this being left for a future enhancement?
2. Where should errors be caught and handled? I notice we call `await mailgun.sendEmail` and don't check the result or catch any error within the action handler. But this seems to be the normal pattern, based on existing action handlers. I'm just curious where in the call stack we catch errors thrown in action handlers?

grahammcculloch (approved):

PR #3288, opened 2020-03-10
Fix meta relationship on org card

ljewalsh:
Fixes meta-relationship on the org card so that members link correctly and can be viewed on the ui


grahammcculloch (approved):
LGTM 

(Still worth adding a useful PR description and removing the placeholder text about writing tests, even for a small PR like this one)

PR #3289, opened 2020-03-10
ui-components: New SlideInPanel component

grahammcculloch:
A generic slide-in panel component that can slide in from any direction. Height (for top/bottom direction) or width (for left/right direction) can be specified. Be sure to mark the appropriate parent element with relative position to display the component as desired.

Change-type: minor

***

Example usage (used in demo gif below):
```javascript
<SlideInPanel height="50%" isOpen={showPanel} onClose={() => setPanelOpen(false)} from="bottom">
  <Box px={3}>
    <h1>Slide-in content goes here...</h1>
  </Box>
</SlideInPanel>
```

### Notes
1. Clicking on the darkened background part of the slide-in panel will automatically cause it to close - unless the `disableClickOutsideToClose` prop is set to `true`.
2. This is a **controlled component** - the `isOpen` state must be controlled by a parent component.
3. `height` (top/bottom variants) or `width` (left/right variants) may be specified. Default behaviour is to use as much space as the child requires.

_The animation is a lot smoother in real-life than this gif shows!_
SlideIn

### DIFF
--- a/lib/ui-components/SlideInPanel.jsx
+++ b/lib/ui-components/SlideInPanel.jsx
@@ -1,0 +1,116 @@
+/*
+ * Copyright (C) Balena.io - All Rights Reserved
+ * Unauthorized copying of this file, via any medium is strictly prohibited.
+ * Proprietary and confidential.
+ */
+
+import React from 'react'
+import styled from 'styled-components'
+import {
+	swallowEvent
+} from './services/helpers'
+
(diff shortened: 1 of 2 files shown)


LucianBuzzo (approved):
LGTM :+1: 
Once we're happy with this component, I think it's a perfect candidate to port to rendition

PR #3309, opened 2020-03-12
ui-components: Add prev/next/action buttons to StepsFlow component

grahammcculloch:
Added unit tests to verify behaviour.

Change-type: minor

***
The `Steps` link behaviour is also slightly modified to align with these buttons. Basically you can only progress through the flow if the current step (and, implicitly, previous steps) are complete. 

The following logic is applied
- Previous button is disabled when you're on the first step
- Next button is disabled if the current step is not 'completed'
- Next button is replaced by the action button on the last step
- Action button is disabled if not all the steps are 'completed'
- A step's link is enabled if:
  -  its a previous step OR
  - its not the current step AND all previous steps are not pending

(Ignore the step content in the following screenshots - focus on the buttons!)

Previous and Next buttons:
NextPrev

Previous and Action buttons:
ActionPrev

### DIFF
--- a/lib/ui-components/StepsFlow.jsx
+++ b/lib/ui-components/StepsFlow.jsx
@@ -10,15 +10,25 @@
 import React from 'react'
 import {
 	Box,
+	Button,
 	Heading,
+	Flex,
 	Steps,
 	Step
 } from 'rendition'
+import _ from 'lodash'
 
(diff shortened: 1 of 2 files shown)


iamsolankiamit (approved):

LucianBuzzo (approved):
LGTM :+1:

PR #3322, opened 2020-03-12
outreach: Add more logging to prospect find by email logic

jviotti:
We keep hitting weird cases where we try to insert a prospect into
Outreach, Outreach complains that the email already belongs to another
prospect, but then we query Outreach to find the prospect that has that
e-mail and the result is none.

Hopefully these extra logs will shed some light.

Change-type: patch
See: https://sentry.io/share/issue/300732926e9e4f2fb0aa347d3d9a045d/

***

test suite!

### DIFF
--- a/lib/sync/integrations/outreach.js
+++ b/lib/sync/integrations/outreach.js
@@ -91,6 +91,10 @@ const getProspectByEmail = async (context, actor, errors, baseUrl, emails) => {
 	}
 
 	for (const email of _.castArray(emails)) {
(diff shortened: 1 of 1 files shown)


grahammcculloch (approved):

grahammcculloch (dismissed):
Looks like there's one linting problem here too!

joshbwlng (approved):

LucianBuzzo (approved):

PR #3323, opened 2020-03-12
Improve loading indicator and error message styles

karaxuna:
Change-type: minor

***

test suite!


grahammcculloch (approved):
LGTM - just need to fix that one linting error.

PR #3325, opened 2020-03-12
Server sdk

Ereski:
Rename `lib/sdk` to `lib/client-sdk` and add a barebones `lib/server-sdk` module


grahammcculloch (commented):
These changes look fine to me. I would just strongly advise you to consult with @LucianBuzzo and @jviotti about:
1. what the server-sdk should look like (what it is/what it isn't)
1. how to manage 'migrating' logic from the client-sdk to the server-sdk in a way that doesn't cause friction or confusion for other developers and ensures nothing falls through the cracks (especially, for example, added/changed code in the client-sdk which isn't picked up and implemented in the server-sdk)

PR #3328, opened 2020-03-13
Complete password reset

ljewalsh:
**NOTE:**
This needs to be merged before the request-password-reset pull request OR the base branch should be changed to master

Action and handler for completing the password reset

- [x] should have a prehook that hashes the submitted password
- [x] should check that the user still exists and has not been disabled
- [x] should check that the token is still valid
- [x] updates user record with new password
- [x] should invalidate the reset-password card (set it to inactive by using the remove card func)
- [x] send success to server


grahammcculloch (commented):
I would recommend someone with more backend experience reviews this as well. But I've left a couple small comments anyway.

ffissore (changes requested):

ffissore (approved):

PR #3362, opened 2020-03-18
Add pagination to table component

StefKors:
This PR adds support for updating the data when the table is paginated. There are a couple of things to note:

1. It makes use of Rendition version 14.2.0 because that version adds support for the `onPageChange` callback.
2. The `itemsPerPage` has been set to 25. By default, we load 30 items in a view. Because we can't know how many items in total exist (due to API limitations) we need to have less than 30 items per page for the table pager to function.
3. In the View lens I moved the `loadViewResults` action to it's own function for use in the `setState` callback. This because the state`page` counter in the action and `setPage` used to get out of sync. Moving it to the callback enforces that `loadViewResults` only triggers after the state updates.

Change-type: patch


grahammcculloch (changes requested):
Just spotted a couple minor things. 

Nice job on the refactor to use the `setState` callback - that's a common but hard-to-spot bug!

grahammcculloch (commented):
Code looks good @StefKors. I'm just reluctant to approve it with no tests. An e2e test for pagination would probably be quite slow/long because you'd have to add more than 25 (more than 30 ideally) items in order to check the pagination functionality. But it might be worth it...

grahammcculloch (approved):
I think this is fine - nice update @StefKors.

For future reference though, I think a more comprehensive unit test of pagination would be appropriate - probably at the ViewRenderer level rather than the CardTable. I don't think we currently have any tests that exercise the `ViewRenderer.setPage` functionality to query the API again and load the new page of data when you change pages.

PR #3375, opened 2020-03-20
Add support for caching arbitrary cards

grahammcculloch:
Change-type: patch

***

### Jellyfish App
This PR basically abstracts the `getActorInternal` method to work with any card (`getCardInternal`). Instead of caching 'actors' in Redux, we cache cards, keyed by card type. Actors are just a simple mapping from a user card, using the (memoized) helper function `generateActorFromUserCard`. A new action is exposed (but not used yet): `getCard`, along with a corresponding selector (called... `getCard`)

### chat-widget
To keep things consistent, the `getActor` action calls through to the `getCard` action. There are two main changes:
1. we use the same 'debounce' functionality as the `getCardInternal` method does in the main Jellyfish app
2. cards are cached in a map keyed by card id - rather than in a flat list

## ui-components
`getActorIdFromCard` and `generateActorFromUserCard` have been moved to `services/helpers.js` and are unit-tested in `services/helpers.spec.js`.


karaxuna (approved):
Very nice. LGTM 👍

karaxuna (commented):
It would be great to hear opinion of others, because I have doubts that it became unnecessary complicated. Instead we could implement cache only in `getActor` method since it only gets a single card.

PR #3398, opened 2020-03-24
add request password reset page (and link from login)

ljewalsh:
Adds the request password reset page to the ui. Also adds a link to this page on the login page (Forgot password?)

- Reset-password page should use a similar form as login and it should take an email
- When the user submits an email, the form should send a action-password-reset-request to the api

TO DO:
- [x] Style the forgot password link so it looks nicer


grahammcculloch (changes requested):
Just one important fix to make and one suggested small refactor!

grahammcculloch (approved):

PR #3399, opened 2020-03-24
Remove CODEOWNERS file

LucianBuzzo:
The use of CODEOWNERS is not well defined and generates a lot of noise. At this point the functionality is causing more harm than good, we're better off selectively choosing reviewers.

Change-type: patch


grahammcculloch (approved):
:shipit:

PR #3401, opened 2020-03-24
Typeform: Add translate tests and integration

mbalamat:


mbalamat (commented):
@jviotti how does this schema look? Should I change the test expectations to use this?

jviotti (approved):
Looks good!

PR #3404, opened 2020-03-24
Links in permissions

Ereski:
Currently, `$$links` entries in the permissions defined in role cards are not accepted. Permissions can only be specified for individual cards (i.e. I can only query session cards that have `data.actor` equal to my user id), and adding constraints that span multiple linked cards is not supported (i.e. I can only query session cards that are linked through the 'is owned by' verb to my user card).

This commit enables support for defining `$$links` in permissions.

Closes #3403

Change-type: minor


ljewalsh (commented):
Hey Carol! Can you add description to the PR with context for this? Helps with reviewing from another timezone and is also useful if we look back through changes to remember why we made them :)

PR #3406, opened 2020-03-25
Fix slugify helper methods

grahammcculloch:
Change-type: patch

***

Fixes #3405 

_Note: I know its not ideal to have multiple `slugify` methods floating around our code base as we currently do. If I understood the rules of dependencies between all our various modules and packages I might attempt to refactor them into one place that is referenced from everywhere else. But I'm leaving that for a separate task._

**Before:**
```javascript
slugify('Balena IO') === '-alena-'
slugify('balena ') === 'balena-'
```

**After:**
```javascript
slugify('Balena IO') === 'balena-io'
slugify('balena ') === 'balena'
```

### DIFF
--- a/lib/action-library/handlers/action-create-card.js
+++ b/lib/action-library/handlers/action-create-card.js
@@ -10,6 +10,7 @@ const assert = require('../../assert')
 
 const slugify = (string) => {
 	return string
+		.trim()
 		.toLowerCase()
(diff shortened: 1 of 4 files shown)


ljewalsh (approved):

jimsynz (approved):
that'll work

PR #3409, opened 2020-03-25
Use filters component from rendition

LucianBuzzo:
Change-type: patch

***

test suite!


StefKors (approved):
Nice! glad we are moving the filters to rendition

PR #3413, opened 2020-03-25
Add GraphQL query interface to our API.

jimsynz:
Add GraphQL server with generated types.

Includes:
* Adds the `graphql.js` GraphQL server.
* Adds schema generation based off our card data model.
* Basic querying by ID and JSON-Schema filter.

Doesn't include:
* Correct resolution of all type fields.
* Implementation of Relay conventions for list fields.
* Query resolution of any but the root fields.

This can be safely deployed because:
1. It doesn't give access to any information you can't already access via the API (it maintains existing session-based security).
2. No one knows it even exists.

Keen for any feedback you may have.

[Here's the spec](https://docs.google.com/document/d/1gP5kZIIxlJxkzGLiR78SzKRM7_gAQernl_LGdNDw4xg/edit?pli=1#).


LucianBuzzo (commented):
Is it possible to use this PR to run HTTP queries yet? I experimented with it, but I only ever got this error:
```
{
  "errors": [
    {
      "message": "Names must match /^[_a-zA-Z][_a-zA-Z0-9]*$/ but \"\" does not."
    },
    {
      "message": "Names must match /^[_a-zA-Z][_a-zA-Z0-9]*$/ but \"30M\" does not."
    },
    {
      "message": "Names must match /^[_a-zA-Z][_a-zA-Z0-9]*$/ but \"2HR\" does not."
    },
    {
      "message": "Names must match /^[_a-zA-Z][_a-zA-Z0-9]*$/ but \"4HR\" does not."
    }
  ]
}
```

Ereski (commented):
A couple of comments in the JSON schema -> GraphQL schema part. There are infinite ways to break it, but I guess it's ok since it's internal-only?

```
allOf: [
  {
    type: 'string'
  },
  {
    format: 'email'
  }
]
```

ljewalsh (approved):
I'm happy with this. From what I can see the concerns are on small details of the code which I think we can address as issues and further PRs

PR #3425, opened 2020-03-27
Improve test init script

joshbwlng:
Change-type: patch

***

Still in testing phase, but should improve maintainability and save time by aborting pipelines more quickly when tests stop responding. 

- Change init script from bash to node
- Run e2e tests sequentially to help reduce test timeouts
- Monitors job output and restarts/kills jobs if no new output within timeout
- Timeout (in seconds) and retry limit may be set with env vars
  - Timeout default = 120 seconds
  - Retry limit = 3


LucianBuzzo (approved):
Looks good to me, If you could address the PR comments in a follow-up point

PR #3426, opened 2020-03-27
Use private SDK package from npm instead of local module

LucianBuzzo:
Connects-to: #3346
Change-type: patch

***

test suite!


StefKors (approved):
LGTM 🚀

StefKors (approved):

PR #3431, opened 2020-03-29
ui-components: new re-usable LinksProvider component

grahammcculloch:
Can be used with matching withLinks and withLink higher order components, or useLink and useLinks hooks.

Change-type: minor

***

In some situations (e.g. dealing with card ownership and the guided handover flow) we need to maintain a cache of linked cards that can be accessed and updated by different child components within a certain context. This is where the generic `LinksProvider` component and its associated HOCs and hooks come in.

Example usage:

```javascript
const SomeComponent = ({ card }) => {
  const { cardOwner, updateCardOwner } = useLink('is owned by', 'cardOwner', card.id)
  return <div>...</div>
}
```

```javascript
<LinksProvider sdk={sdk} cards={[ card ]} link="is owned by">
  ...
  <SomeComponent card={card} />
</LinkProvider>
```

### DIFF
--- a/lib/sdk/link-constraints.js
+++ b/lib/sdk/link-constraints.js
@@ -3,6 +3,18 @@
  * Unauthorized copying of this file, via any medium is strictly prohibited.
  * Proprietary and confidential.
  */
+const _ = require('lodash')
+
+exports.supportsLink = _.memoize((cardType, linkName) => {
+	return Boolean(_.find(
+		exports.constraints,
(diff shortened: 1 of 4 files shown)


karaxuna (changes requested):
Additional question: why can't we store links inside redux store?

ljewalsh (changes requested):

ljewalsh (approved):

karaxuna (approved):

PR #3432, opened 2020-03-29
ui: Add support for flows in the redux store

grahammcculloch:
Change-type: patch

***

In preparation for the addition of the first of various 'guided flows', this PR just adds support for managing flow state in the redux store.

_Note: these action creators are not called by any code yet. That will come in a following PR!_


ljewalsh (approved):
LGTM!

PR #3438, opened 2020-03-30
Add new lens for creating a user

ljewalsh:
**NOTE:** future work will allow the operator to send first-time-login-links to existing users
**Another note:** I've moved the actioncreators into their own folder so we can start breaking the tests down into smaller files. I havent done that work in this PR because I didnt want to complicate things

This allows users with the operator role to create users. Essentially, it introduces a createUserLens with a email and username input.

On submit, the client makes a create-user request with the email, username, and an empty string as a password. I've updated the backend so that it sets the password to the PASSWORDLESS_USER_HASH when an empty password is sent to the create-user action (later on in this process we expect the user to have no password set yet).

After the user is created, the client fires the createLink redux to create an org link (and therefore attaches the created user to the operator-user's org). It then fires the send-first-time-login-link action so that a first-time login link is sent to the supplied email address.

Rather than having to copy and paste a hash to the user over flowdock, we can now use this lens to create a user, send them a first-time-login link and let them set their password themselves.

TO DO
- [x] add tests


grahammcculloch (changes requested):

grahammcculloch (changes requested):
Just a few things to tidy up.

grahammcculloch (changes requested):
Just found one significant thing to fix (see the comment about `data.types`)

grahammcculloch (changes requested):
Just a tiny (but important!) correction to your org slug regex to make.

grahammcculloch (approved):

PR #3442, opened 2020-03-30
Generate CardTable table columns from card schema

StefKors:
Fixes: #3440
Change-type: patch

Table columns will be generated based on the schema properties from the type card used. The column name used will be the title of the column property, if no title is present the flattened property path will be used.

Screenshot 2020-04-01 at 13 42 20

***

test suite!


grahammcculloch (changes requested):
I like that you've done the 'heavy lifting' in a separate `helpers.js` file. I would love to see a `helpers.spec.js` file with unit tests for these helper methods. This would make it much clearer to the reader what the functions do and would let you test out the corner cases that normal use on the site might not reach.

grahammcculloch (approved):

PR #3443, opened 2020-03-30
Fixes request-password-reset and complete-password-reset

ljewalsh:
Several problems found with the reset process:

- We were trying to get the user by email with the guest session which meant we could not see the hash on the user
- We had the password_reset url around the wrong way
- We were passing the wrong card to the patch


grahammcculloch (changes requested):
Are you able to test this in an e2e/integration test? 

I guess if you can verify the bug and the fix when manually testing locally that should be good enough.

grahammcculloch (approved):

PR #3445, opened 2020-03-31
Use plain subqueries instead of CTEs when querying for links

Ereski:
This should reduce DB load when running queries that contain a `$$links`
property

Change-type: patch

(fixing unit tests for `jsonschema2sql`)


petrosagg (approved):

jimsynz (approved):
Approved, pending tests.

do it now!

jviotti (approved):

PR #3446, opened 2020-03-31
Revert "Use plain subqueries instead of CTEs when querying for links"

Ereski:
This reverts commit 46ab4123d09b74522acad9a570d9198dab7ad932.

The DB is not behaving as expected


petrosagg (approved):

jimsynz (approved):
revert!

PR #3450, opened 2020-04-01
Automatically create partial indexes based on type cards

LucianBuzzo:
A new `indexed_fields` field is added to type cards, this field's value
is an array of arrays of strings. Each sub-array of strings represents
columns in the database. If the column to be indexed is a jsonb column,
it is represented using a '.' delimiter.
On create or update, if a type card has an `indexed_fields` property,
partial indexes will be created for that type using the columns named.

For example, if the link type card has `indexed_fields` set to:
```
[
	[ 'data.from.id', 'name', 'data.to.id' ]
]
```
A single multi-column partial index will be created, constrained to
cards where the type is "link".

Change-type: minor

***

test suite!

### DIFF
--- a/lib/core/backend/postgres/cards.js
+++ b/lib/core/backend/postgres/cards.js
@@ -6,6 +6,7 @@
 
 const _ = require('lodash')
 const Bluebird = require('bluebird')
+const pgFormat = require('pg-format')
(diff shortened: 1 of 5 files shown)


jviotti (approved):
LGTM

PR #3451, opened 2020-04-02
Adds more warning logs to request-password-reset

ljewalsh:
Currently the request-password-reset action works locally but not on production. No logs seem to be coming through which suggests that the problem is with the data returned by getUserByEmail. Sometimes permission issues means we find the user but there is no hash attached. I've added a bunch of extra internal logs to help debug this problem


jimsynz (approved):
awesome

PR #3458, opened 2020-04-02
Bring back $$formula type support

jviotti:
***

test suite!


StefKors (changes requested):
@jviotti  I'm trying to get a simple SUM example running. I've added a commit with a test for it but it doesn't work... Could you point me in the right direction for using other fields in $$formula?

StefKors (approved):

PR #3460, opened 2020-04-02
github: Setup GitHub App id and private key env vars

jviotti:
These env vars will be used by the GitHub integration to make use of
GitHub Apps when possible.

Change-type: minor

***

test suite!


mbalamat (approved):

ljewalsh (changes requested):
I thnk you will also want to put these in the docker files Juan

ljewalsh (approved):

PR #3461, opened 2020-04-02
github: Instantiate different Octokit clients for translate/mirror

jviotti:
We can't have a single GitHub Octokit client for the whole integration
once we move to GitHub Apps.

This commit prepares the integration so that we instantiate Octokit
clients on the fly on the `.translate()` and `.mirror()` methods.

Change-type: patch

***

test suite!


ljewalsh (approved):
This looks good to me. Happy for it to be merged once the tests pass

PR #3465, opened 2020-04-03
ui-components: Small fix for LinksProvider

grahammcculloch:
Handles the situation where cards is null.

Change-type: patch

***


StefKors (approved):
:shipit:

PR #3473, opened 2020-04-06
Update request password reset action and ui to take username instead of email

ljewalsh:
We made the assumption that emails are unique to users when creating the request-password-reset functionality. Unfortunately this is not true and because of syncing there are often multiple users under our Balena emails.

This is technical debt we should probably fix but in the meantime we want people to be able to reset their passwords. So I have shifted our UI and backend to use the username instead

### DIFF
--- a/apps/ui/core/store/actioncreators.js
+++ b/apps/ui/core/store/actioncreators.js
@@ -894,7 +894,7 @@ export default class ActionCreator {
 	}
 
 	requestPasswordReset ({
-		email
+		username
 	}) {
 		return async (dispatch, getState) => {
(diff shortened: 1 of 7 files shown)


grahammcculloch (commented):
Would it not make more sense to just _add_ the username field to the password reset code rather than _replace_ the email field with the username field everywhere? Yes - it's arguably redundant info, providing both username and email, but this way it will be _much_ easier to just strip out the username references once we resolve the technical debt and revert to the user just having to provide the email address. 

And it will be slightly less weird for users who are used to being asked for their email address when requesting a password reset.

grahammcculloch (approved):

PR #3492, opened 2020-04-08
Add send-first-time-login-link and handler to action-library

ljewalsh:
At the moment, when a user is created in Jellyfish, we have to go into the db at setup a temporary hash and send this to the user over flowdock. We then ask them to reset their password once they have logged in.

This action will be available to users with the operator-role so that they can send a one-time-link to the user when they are first created. This way they no longer need to send a temporary password hash over flowdock

The handle should:
- [x]     Confirm that the user belongs to the same organisation as the requester
- [x]     Confirm that the requester has an operator role
- [x]     Confirm that the user is active (not disabled)

### DIFF
--- a/apps/server/card-loader.js
+++ b/apps/server/card-loader.js
@@ -59,6 +59,7 @@ module.exports = async (context, jellyfish, worker, session) => {
 		await loadCard('contrib/feedback-item.json'),
 		await loadCard('contrib/issue.json'),
 		await loadCard('contrib/message.json'),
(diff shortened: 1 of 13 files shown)


LucianBuzzo (approved):
LGTM, a few nitpicks though: 
- In the future, please split out refactors into a separate commit
- Please try and make your commit title more meaningful, something like "Add ability to send first-time login links to new users" would be much clearer

PR #3506, opened 2020-04-13
core: Mention explicit columns in Postgres adapter getters

jviotti:
This gives us more control over the columns we return to make it easier
to add extra columns in a backwards compatible way.

Change-type: patch

***

test suite!


joshbwlng (approved):

ljewalsh (approved):
LGTM

PR #3512, opened 2020-04-14
Remove ability for community users to add users

ljewalsh:
We want to remove the ability for community users to be able to create other users and restrict this to users who also have the operator role. There seems to be a bug in the system. I've updated the test to expect a fail but it passes


jimsynz (approved):
Bonus points for replacing the HTTP calls with SDK calls.

You're A King To Me

PR #3524, opened 2020-04-17
Add Prometheus ports/paths to deploy templates

joshbwlng:
Change-type: patch

***

Add Prometheus metrics ports and paths to production deploy templates.


cmfcruz (approved):
Let's try this out. It shouldn't break anything.

PR #3556, opened 2020-04-22
core: Use major/minor/patch version columns in queries

jviotti:
Change-type: minor
Signed-off-by: Juan Cruz Viotti <juan@balena.io>


StefKors (approved):
LGTM

PR #3557, opened 2020-04-22
Fix a number of bugs preventing us from using Jellyfish as a flowdock/slack replacement

LucianBuzzo:
***

test suite!


grahammcculloch (commented):
Wow - I really like that 'invert-the-query-with-a-mask' trick!

Just spotted a couple of stray `console.log` calls that should probably be removed.

jviotti (approved):

PR #3598, opened 2020-04-27
UI updates

LucianBuzzo:
Bundles a number of UI related bug fixes.
Each fix has a separate commit, so take a look through this PR commit-by-commit for additional detail

- Fix deprecated uuid imports
- Fix bug where cards would be requested despite being in cache
- Don't unnecessarily make requests for cards that are already cached
- Fix bug that stops view slices from working correctly
- Make the UI more resilient to network disruption


grahammcculloch (approved):
LGTM (although don't fully understand the view slices fix - is that verified in a test?)

PR #3600, opened 2020-04-27
Fix front urls in messages

StefKors:
Instead of pre-fixing only the first URL in a message, this change will match and pre-fix all Front URLs in a message.

Fixes: #3012
Change-type: patch

***

test suite!


ljewalsh (approved):
Looks great

PR #3605, opened 2020-04-28
Fix metrics integration tests

joshbwlng:
Change-type: patch

***

Metrics tests were a bit broken, especially when running against a locally running Jellyfish.
Some logic was assuming tests were to only be ran in CI/compose, which is not OK for our current situation.

Examples of how to run metrics tests against a local Jellyfish:
```
# start local Jellyfish
$ make start-postgres
$ make start-redis
$ METRICS_PORT=9200 make start-server
$ make start-tick
$ METRICS_PORT=9300 make start-worker
$ make dev-livechat
$ make dev-ui

# run metrics tests
$ METRICS_PORT=9400 make test-integration-worker
$ METRICS_PORT=9400 SOCKET_METRICS_PORT=9500 SERVER_PORT=8100 make test-integration-server
```

Nothing should change for CI/compose-based tests.


StefKors (changes requested):
Nice fix! let's also add a quick update to the docs here: 
https://github.com/product-os/jellyfish/blob/master/docs/developing/running-tests.markdown

StefKors (approved):

PR #3613, opened 2020-04-28
Add a reworked `jsonschema2sql` under a feature gate

Ereski:
This rework is necessary to support further improvements, some of which are already incorporated in this commit such as:

- A new query structure for linked cards relying on plain joins instead of subqueries. Benchmarks show that this new structure is consistently faster than the old query structure even on difficult edge cases;
- Better error checking instead of silently ignoring wrong schemas;
- Correct query generation for some nested keywords (i.e. nested `required` that is currently ignored);
- Other minor improvements and handling of rare corner cases.

This PR also prunes:

- The experimental alternative query generator that supports `$$links` within `$$links`. This code path is unused in production and conflicts with this rewrite. If needed, this feature can be reintroduced in the future;
- Support for setting `null` as the schema for a linked cards, signifying that the link must not exist. This functionality is not currently used and would be obsoleted anyway when support for nested links lands;
- Support for the following keywords (may be readded later):
  - `default`
  - `patternProperties`
  - `propertyNames`
  - `formatMaximum`
  - `formatMinimum`

Change-type: minor


ljewalsh (dismissed):
Hey @Ereski! I dived into this for a while but got distracted trying to setup a debugger on vim so I could play around with it easier. Here are just a couple of comments. I'd like to have had more for you but it will have to wait for tomorrow

LucianBuzzo (changes requested):
Out of curiosity, does the suffix `Visitor` have a special meaning outside of this work?

petrosagg (approved):

LucianBuzzo (approved):

PR #3616, opened 2020-04-29
backend: Add initial full text search support

joshbwlng:
Change-type: patch

***

This PR:
- builds `tsvector` partial indexes on init for each card field that has `fullTextSearch: true`
  - Indexes for `text[]` columns are created using a new custom function `immutable_array_to_string()` in order to achieve the same `tsvector` results
  - Indexes for values that exist under `data` are created using `jsonb_to_tsvector()`
  - Indexes for `text` columns are created using `to_tsvector()`
- adds `fullTextSearch` keyword handlers to the `jsonschema2sql` compiler
- adds full text search functionality, initially for these cards:
  - `support-thread`
    - `name`, `tags`, `data.category`, `data.tags`, `data.environment`, `data.description`, `data.inbox`, `data.statusDescription`
  - `sales-thread`
    - `name`, `tags`, `data.tags`, `data.inbox`, `data.statusDescription`, `data.status`
  - `thread`
    - `data.tags`, `data.description`
  - `message`
    - `data.timestamp`, `data.payload.file.name`, `data.payload.message`
  - `whisper`
    - `data.timestamp`, `data.payload.file.name`, `data.payload.message`

***

Full-text search using `tsvector`s will allow to perform searches more efficiently by reducing terms to their stems (e.g. `cats` => `cat`) and removing "noise" that will not help in search (e.g. a, and, the).

Here are a couple examples of what some text looks like after being transformed into `tsvector`s using [`to_tsvector`](https://www.postgresql.org/docs/9.1/textsearch-controls.html):
```
# SELECT to_tsvector('english', 'cats and dogs');
   to_tsvector
-----------------
 'cat':1 'dog':3

# SELECT to_tsvector('english', 'Add full-text search support for strings nested inside of complex structures');
                                                       to_tsvector
--------------------------------------------------------------------------------------------------------------------------
 'add':1 'complex':12 'full':3 'full-text':2 'insid':10 'nest':9 'search':5 'string':8 'structur':13 'support':6 'text':4
```

These `tsvector`s are stored in partial indexes, one index per field per card type. For example, a partial index for the `data.payload.message` field on the `message` card will be created as `message__data_payload_message__search_idx`.

Search queries are prepared to be ran against these indexes using [`plainto_tsquery`](https://www.postgresql.org/docs/9.1/textsearch-controls.html): 
```
SELECT slug FROM cards WHERE to_tsvector('english', cards.data#>'{description}') @@ plainto_tsquery('english', 'job explore') AND type='support-thread@1.0.0';
```

***

Things to be done in subsequent PRs:
- Add full-text search support for strings nested inside of complex structures (object->array->object->string) such as `data.payload.attachments.name` in `message` cards

Closes https://github.com/product-os/jellyfish/issues/386

### DIFF
--- a/apps/server/default-cards/contrib/message.json
+++ b/apps/server/default-cards/contrib/message.json
@@ -21,7 +21,8 @@
           "properties": {
             "timestamp": {
               "type": "string",
-              "format": "date-time"
+              "format": "date-time",
+              "fullTextSearch": true
             },
             "actor": {
               "type": "string",
@@ -47,7 +48,8 @@
                   "type": "object",
                   "properties": {
                     "name": {
-                      "type": "string"
+                      "type": "string",
+                      "fullTextSearch": true
                     },
                     "mime": {
                       "type": "string"
@@ -83,7 +85,8 @@
                 },
                 "message": {
                   "type": "string",
-                  "format": "markdown"
+                  "format": "markdown",
+                  "fullTextSearch": true
                 }
               }
             },
--- a/apps/server/default-cards/contrib/sales-thread.json
+++ b/apps/server/default-cards/contrib/sales-thread.json
@@ -12,13 +12,15 @@
       "type": "object",
       "properties": {
         "name": {
-          "type": [ "string", "null" ]
+          "type": [ "string", "null" ],
+          "fullTextSearch": true
         },
         "tags": {
           "type": "array",
           "items": {
             "type": "string"
-          }
+          },
+          "fullTextSearch": true
         },
         "data": {
           "type": "object",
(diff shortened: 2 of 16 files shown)


grahammcculloch (commented):
A few more questions for you Josh :-)

Also - have you tried a test that does a single full text search on message _and_ whisper cards? Does it work?

**Reply from Josh:** Don't have that test written up at the moment, but will look at adding one. I'm not modifying any core SQL generation logic so it should work like current `regexp` searches do, but I'll double-check.

Ereski (changes requested):
A lot of comments, mostly related to handling of types.

Ereski (dismissed):

Ereski (approved):

PR #3622, opened 2020-04-29
Display timeline images inside MessageContainer component

StefKors:
With this PR uploaded images show what type of message they are.

Change-type: patch

Screenshot 2020-04-30 at 12 35 57

***

test suite!


grahammcculloch (changes requested):
Couple things that need fixing I think.

grahammcculloch (approved):

PR #3676, opened 2020-05-06
ui: Use react router for internal links

StefKors:
Fixes: https://github.com/product-os/jellyfish/issues/2290

This PR replaces the use of the rendition `Link` component in Jellyfish with the `RouterLink`component. RouterLink wraps the rendition component with the following behavior:

- Relative links use react-router for navigation.
- External links use default browser behavior.
- Links that navigate to the current `window.location` are treated as relative links.

### DIFF
--- a/apps/ui/lens/full/MyUser/MyUser.jsx
+++ b/apps/ui/lens/full/MyUser/MyUser.jsx
@@ -17,8 +17,7 @@ import {
 	Txt,
 	Tab,
 	Tabs,
-	Divider,
-	Link
+	Divider
 } from 'rendition'
 import {
 	Form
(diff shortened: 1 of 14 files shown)


grahammcculloch (changes requested):
@StefKors I fixed up a few things and squashed my updates into your first commit. Here's the highlights:

1. `Link.navigate()` returns true if the link is not going to be handled by `history` - see [this explanation](https://stackoverflow.com/questions/1346043/html-anchor-link-href-and-onclick-both)
1. Removed un-associated changes to `package-lock.json`
1. Refactored the `blank` parameter out of `isExternalUrl` as this conceptually  unrelated. We still obviously check for whether `blank` is specified when navigating.
1. Tidied up the unit test methods to make them more readable/concise.

I think there's one thing missing: if the link url starts with `window.location.origin`, I think we should strip that out (i.e. convert it to a relative url). This way we avoid a whole page refresh on absolute urls that are on the same site.

PR #3702, opened 2020-05-07
Refactor inbox to use materialized field instead of regexp

LucianBuzzo:
Previously the inbox had been disabled due to performance and stability
concerns. A chunk of this PR is simply re-enabling the inbox
functionality that as turned off in this PR https://github.com/product-os/jellyfish/pull/3568

Change-type: minor

***

test suite!

### DIFF
--- a/apps/server/default-cards/contrib/view-my-inbox.json
+++ b/apps/server/default-cards/contrib/view-my-inbox.json
@@ -10,7 +10,7 @@
   "data": {
     "allOf": [
       {
-        "name": "message or whisper",
+        "name": "My unread pings",
         "schema": {
           "type": "object",
           "properties": {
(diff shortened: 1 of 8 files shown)


StefKors (commented):
Not sure if this should be fixed in this PR or in a new issue. 

Testing this PR I found out that when clicking the notification, it opens the channel but it doesn't modify the URL. More importantly, this makes it impossible to close the channel opened by a notification.

I'm taking a look at fixing this as it seems to be related to the generic `ADD_CHANNEL` store action. Hinting at that it would impact everything created by this action.

grahammcculloch (commented):
@LucianBuzzo I had a play with this branch locally and noticed a few issues to do with the search logic. I've pushed a commit that fixes these issues. If you're happy with that, ping me and I'll approve this PR.

grahammcculloch (approved):

PR #3708, opened 2020-05-07
Sort results of the new and old compilers

Ereski:
Avoid spurious mismatches

Change-type: patch

***

test suite!


grahammcculloch (commented):
@Ereski I'm surprised that `results` and `newResults` aren't already sorted the same. Presumably we want `newResults` to exactly match `results` without having to be sorted by card id first?

grahammcculloch (approved):

PR #3724, opened 2020-05-11
ci: Restore CircleCI Postgres server dump

joshbwlng:
Change-type: patch

***

We are seeing consistent failures on https://github.com/product-os/jellyfish/pull/3699 when attempting to execute the `e2e_tests_server_previous_dump` job. After doing a bit of debugging I was able to confirm that the `curl` call to the CircleCI API to retrieve previous run artifact information resulted in an empty array. I believe this is caused by the fact that we have not been storing Postgres dumps as artifacts since https://github.com/product-os/jellyfish/commit/94683cabe0f8f569efcf82e9fe814657457c39a8

This PR will begin storing Postgres dumps as artifacts again allowing subsequent `e2e_tests_server_previous_dump` jobs to download previous dumps and run properly.


grahammcculloch (commented):
@joshbwlng would it be valid just to change the `e2e_tests_server_previous_dump` job to handle an empty array? Presumably if the curl call _failed_ it wouldn't return an empty array, it would return something else (and an HTTP error code)?

grahammcculloch (approved):

PR #3726, opened 2020-05-11
environment: Trim Outreach client id and client secret env vars

jviotti:
Looks like we have a trailing whitespace that is confusing
https://api.outreach.io/oauth/token in production.

Change-type: patch

***

test suite!


LucianBuzzo (approved):
Should we be doing this for every env var?

PR #3730, opened 2020-05-11
Allow users to have multi-threaded conversation about a repository

LucianBuzzo:
This change allows `repository` type cards to have multiple threads attached to them, these threads are then rendered using the interleaved lens. This behaviour lets repos behave like "chatrooms" in Jellyfish, allowing us to organise our work discussion around the code we write and repos we work on.

image

Change-type: patch

***

test suite!


karaxuna (approved):
:+1:

PR #3737, opened 2020-05-12
ui: Make tab headers horizontally scrollable on small screens

grahammcculloch:
And force the tab header text to stick to a single line.

Change-type: patch

***

This is a proposed change to how we render tab headers on small screens. Let me know what you think. There's pros and cons but it definitely takes up less vertical screen space on small screens which is a big plus!

Note - the behaviour of tab headers on larger screens is unchanged.

Before:
Tabs - wrapped

After:
Tabs - single line

### DIFF
--- a/apps/ui/index.jsx
+++ b/apps/ui/index.jsx
@@ -74,6 +74,23 @@ const customTheme = {
 		},
 		background: '#fff',
 		border: '#eee'
+	},
+	tabs: {
+		header: {
+			extend: (props) => {
+				return `
+					@media only screen and (max-width: ${props.theme.breakpoints[1]}px) {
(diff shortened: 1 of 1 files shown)


StefKors (changes requested):
@grahammcculloch Iove the improvement, 

1) Rendition could benefit from this as well let's PR it there instead. I think adding it through the `compact` breakpoints or `renderMode` prop might be a good idea.

2) Let's also improve the vertical height of this tab component. The 39px vertical height of the tabs is just a bit too small for mobile use, including a bit more padding on top could make it mobile-friendly without having to reconfigure the font-size or vertical spacing too much.

PR #3751, opened 2020-05-12
Avoid recalculating `SqlQuery.isProcessing*()`

Ereski:
This results in a minor performance boost and unlocks a nicer
implementation of PR #3716.

Change-type: patch

***

test suite!


jimsynz (approved):
:shipit:

PR #3752, opened 2020-05-12
Don't use `_.defaults()` for better performance in `SqlQuery` constructor

Ereski:
Profiling showed that lodash's `defaults()` was taking a not
insignificant amount of time. Benchmarking showed that using plain `in`
operators to set the default values led to a not insignificant
performance boost.

Change-type: minor

***

test suite!


jimsynz (approved):
:shipit:

PR #3758, opened 2020-05-13
ui: Navigate to default view if set when home page loaded

grahammcculloch:
Change-type: minor

***

Closes #3712 

This PR finishes off some functionality which seemed to be partially implemented. You can now set a particular view as the default view to load when logging in (or loading the home page). The default view has a little 🏠 icon next to it. You can also now 'unset' the default view - to reset it to normal behaviour.

DefaultView


StefKors (changes requested):
Looks great @grahammcculloch 
Nit: Could you add a test for this?

StefKors (approved):

PR #3762, opened 2020-05-13
Reduce emission of type-checking conditionals in the new compiler

Ereski:
Many JSON schema keywords must be implemented with a material conditional at the SQL level, equivalent to:
    
```
typeof <field> === <type> ? <condition> : true
```
    
This is implemented as a boolean expression at the SQL level:
    
```
NOT <isType> OR <condition>
```
    
where `<isType>` is the equivalent SQL for the `typeof` check. Frequently the type is specified in the schema, leading to the following SQL expression:
    
```
<isType> AND (NOT <isType> OR <condition>)
```
    
which siplifies to just `<condition>`. PG does not recognize this optimization opportunity so this commit applies it at the compiler level.
    
Change-type: patch

### DIFF
--- a/lib/core/backend/postgres/jsonschema2sql/index.js
+++ b/lib/core/backend/postgres/jsonschema2sql/index.js
@@ -1198,7 +1198,7 @@ COALESCE(${table}.version_patch, '0')::text
 		if (typeFilter === null) {
 			return filter
 		} else if (typeFilter === false) {
-			return new SqlFilter(false)
+			return new SqlFilter(true)
 		}
 
 		return SqlFilter.materialConditional(typeFilter, filter)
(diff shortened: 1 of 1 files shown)


grahammcculloch (changes requested):
@Ereski this is a large change (in terms of modified lines of code) in a core part of the code with no tests added for it. Is it possible to add some unit tests/integration tests to prove (and document) the changes in this PR? I would be nervous to approve it without the reassurance that comes from having some tests in place!

BTW this PR description was perfect - suuuuper clear explanation of a logical nuance. Really appreciate the time taken to give a clear explanation!

jviotti (approved):
@Ereski Looks good but there are some conflicts we need to resolve.

grahammcculloch (approved):

Issue #3651, opened 2020-05-05
Use Service Worker for push notifications


grahammcculloch:
As we're converting Jellyfish into a Progressive Web App, we have the opportunity to show users push notifications via the service worker - which means they can get notifications even when they don't have the app open!

## Proposed Plan

The [`web-push`](https://github.com/web-push-libs/web-push) npm package provides a lot of the required functionality.

### Pre-requisites

1. First we need a set of [VAPID](https://stackoverflow.com/a/40399203/11480470) keys (public and private). We can use [`webpush.generateVAPIDKeys()`](https://github.com/web-push-libs/web-push#generatevapidkeys) to create these. The frontend needs the public key (as an env var) and the server needs both private and public keys (again as env vars).
2. For better compatibility we should also generate and use a Google Cloud Messaging (GCM) sender id and API key. See [this article](https://www.izooto.com/blog/generate-gcm-sender-id-server-api-key) for details. The GCM sender id is added to the `manifest.json` and the API key is used by the server when sending notifications (stored as an env var).

### Steps

1. When the Home Channel is mounted, the frontend checks if notifications permission has been granted. If not, we ask for permission. If granted*:
    1. we call [`serviceWorkerRegistration.pushManager.subscribe()`](https://developer.mozilla.org/en-US/docs/Web/API/PushManager/subscribe) 
    2. and then use the subscription details to create and save a new `web-push-subscription` card with the following fields:
        * `userId` (represented as a link?)
        * `endpoint` - (unique/primary key) Push Subscription URL
        * `token` - User Public Encryption Key
        * `auth` - User Auth Secret
2. The frontend [listens for changes to notification preferences](https://stackoverflow.com/a/39880338/11480470). If it detects a change, it subscribes or unsubscribes the client as necessary. (Subscribing also includes upserting an appropriate `push-subscription` card. Unsubscribing includes [getting the existing subscription](https://developer.mozilla.org/en-US/docs/Web/API/PushManager/getSubscription), calling [`unsubscribe()`](https://developer.mozilla.org/en-US/docs/Web/API/PushSubscription/unsubscribe) on it and deleting the matching `push-subscription` card.)
[...]

PR #3777, opened 2020-05-15
server: Add new web-push-subscription type card

grahammcculloch:
Users with the user-community role can view their own web-push-subscription cards.

Change-type: minor

***

Relates to: #3651 

Pre-requisite for implementing Service Worker Push Notifications.

**TBD**: Are we ok to store the `auth` value 'as-is' or do we need to encrypt it in some way that only the server can decrypt?


jimsynz (approved):
you have my axe!

joshbwlng (approved):

PR #3794, opened 2020-05-18
Use speech-bubble icon for the first thread message in the repo channel

ljewalsh:
At the moment, all messages in a repository have the same icon next to them, making it hard to differentiate where one thread ends and another starts.

This PR introduces a temporary solution to replace the usual icon with a speech bubble icon when the message is the first in a thread. This makes our channel look a little more like Flowdock.

Before:

Screen Shot 2020-05-18 at 1 27 15 PM

After: 

Screen Shot 2020-05-18 at 1 21 24 PM

- [x] Ensure messages are date-sorted
- [x] Move first-message testing to Interleave component
- [x] Ensure other types of linked card are ignored (e.g events)
- [x] fix bug with view-all-messages loading animation
- [x] test various situations in which interleave is used

### DIFF
--- a/apps/ui/lens/list/Interleaved.jsx
+++ b/apps/ui/lens/list/Interleaved.jsx
@@ -50,6 +50,17 @@ const isHiddenEventType = (type) => {
 	return _.includes(NONE_MESSAGE_TIMELINE_TYPES, type)
 }
 
+// TODO: remove once we can retrieve this data during query
+const isFirstInThread = (card, firstMessagesByThreads) => {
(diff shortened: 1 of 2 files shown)


grahammcculloch (changes requested):
I'm still not really sure about this approach overall so lets leave this as a draft PR. In the meantime, I've pointed out some places to optimise the code.

grahammcculloch (changes requested):

grahammcculloch (approved):

PR #3806, opened 2020-05-18
On the interleaved lens, close open threads when opening a new thread

LucianBuzzo:
Fixes #3694

Change-type: patch

***

test suite!


grahammcculloch (commented):
@lucianbuzzo is `current` always an ID? If so, will this fail when the URL uses the view `slug`?

grahammcculloch (approved):

PR #3815, opened 2020-05-19
Trim whitespace from AWS S3 access credentials

LucianBuzzo:
This change help defend against situations where env vars may have extra
whitespace in them.

Change-type: patch
Connects-to: #3706

***

test suite!

### DIFF
--- a/lib/environment/index.js
+++ b/lib/environment/index.js
@@ -16,6 +16,8 @@
  *   console.log('Production!')
  * }
  */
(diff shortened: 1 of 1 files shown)


StefKors (commented):
@LucianBuzzo what is the advantage of using lodash `_.trim()` vs the plain js `.trim()`?

joshbwlng (dismissed):

joshbwlng (approved):

PR #3826, opened 2020-05-20
ui-components: Fix textarea autosizing bug

grahammcculloch:
Replaced the package react-autosize-textarea with the more popular package react-textarea-autosize - which is compatible with @webscopeio/react-textarea-autocomplete.

Change-type: patch

***

Fixes #3825 

Recently we broke the textarea autosizing by [bumping react-autosize-textarea from v6 to v7](https://github.com/product-os/jellyfish/pull/3748). 

It turns out there's an incompatibility between `react-textarea-autocomplete` and `react-autosize-textarea` - see [this issue](https://github.com/webscopeio/react-textarea-autocomplete/issues/150).

We could just downgrade back to v6 of `react-autosize-textarea` - but that doesn't seem like a good longterm plan! Instead, I've switched out the dependency to the [more popular](https://www.npmtrends.com/react-autosize-textarea-vs-react-textarea-autosize) [`react-textarea-autosize`](https://github.com/andreypopp/react-textarea-autosize) (annoyingly similar name!) which works just fine with `react-textarea-autocomplete`.


joshbwlng (approved):

ljewalsh (approved):
Thanks for fixing this up @grahammcculloch

PR #3827, opened 2020-05-20
Update CardField to display arrays and objects as a HTML list

ljewalsh:
At the moment, when we display the fields on a card, arrays come out looking a little janky with the index of each value as the title:
Screen Shot 2020-05-20 at 5 10 26 PM

This PR updates the CardField component so that we display both objects an arrays as a `<ul>` HTML tag. Each item in the list or key/value in the object are displayed as a`<li>` within the `<ul>`:
Screen Shot 2020-05-26 at 11 52 44 AM

Key/values are displayed as <li>key: value</li>
Screen Shot 2020-05-26 at 11 55 58 AM

Also adds some basic enzyme tests


grahammcculloch (changes requested):
I like the new tests - this will be useful as we extend the `CardField` functionality in the future!

Just one suggested change to account for arrays of items that aren't strings.

karaxuna (changes requested):

LucianBuzzo (changes requested):

grahammcculloch (approved):
Approved - on the understanding that we're going to revisit this component soon to handle different field types (e.g. arrays of objects) in a more comprehensive, testable way.

PR #3836, opened 2020-05-21
Add Google Meet integration

grahammcculloch:
The standard card actions component now includes a video link that will open a Google Meet specific for that card in a new tab/window. If no link exists, one will be created, saved to the card's data and then the link opened.

Change-type: minor

***
Before this PR was merged, we deployed the appropriate secret to the k8s cluster that will populate the following environment variable:

`INTEGRATION_GOOGLE_MEET_CREDENTIALS`

The value of this environment variable is the stringified JSON value of the Google Meet JWT credentials. For example:

```sh
export INTEGRATION_GOOGLE_MEET_CREDENTIALS="{\"type\":\"service_account\",\"project_id\":...
```

For reference, the relevant JSON can be found in the 'notes' section of the `Google Meet JWT` **passpack** item. 

***

Closes #3802 

image

Summary of behaviour:
* Initially, a card has no `data.conferenceUrl` value set. A `VideoLink` is displayed in a lighter tone with the tooltip `Request a Google Meet for this <card-type>`. 
* When clicked, the link will display a spinning cog and the tooltip `Requesting a Google Meet for this <card-type>`.
* Clicking on the link will initiate a `action-google-meet` action.  This will use the Google API to: 
  1. create a calendar event
  1. extract the Google Meet URL from the created event data
  1. delete the event
  1. patch the `data.conferenceUrl` field in the card with the new Google Meet URL
  1. return the Google Meet URL to the client
* When the URL is returned from the server, the front-end opens it in a new tab/window
* Subsequent clicks on the `VideoLink` component will just directly open the Google Meet URL in a new tab/window with no interaction with the server.


LucianBuzzo (dismissed):

cmfcruz (dismissed):
LGTM

joshbwlng (changes requested):

joshbwlng (approved):

PR #3850, opened 2020-05-22
Refactor Event component: extract Header and Body into separate components

ljewalsh:
This is the first step in some larger work to break the Event component down into smaller, more easily testible components.

This PR creates a `<EventHeader />` and `<EventBody />` component and moves the respective code from `<Event />` into these components.

### DIFF
--- a/lib/ui-components/Event/Event.jsx
+++ b/lib/ui-components/Event/Event.jsx
@@ -4,7 +4,6 @@
  * Proprietary and confidential.
  */
 
-import copy from 'copy-to-clipboard'
 import {
 	circularDeepEqual
 } from 'fast-equals'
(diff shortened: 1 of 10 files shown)


grahammcculloch (changes requested):
@ljewalsh I found quite a few things that need fixing (maybe you've fixed some of them in the time it took me to check through all this and write comments?)

grahammcculloch (changes requested):
Looks better @ljewalsh but I still found something missing :-)

grahammcculloch (approved):
LGTM now. 

FWIW one of the main benefits of splitting up a component into smaller components is that you can unit test each smaller component on its own - which should be much easier than 'unit' testing the old top-level component. In this PR you've split up the component into smaller components but then added tests to the top-level component. In the future, when breaking up components like this I would recommend adding unit tests to the new smaller child components that have been created.

PR #3858, opened 2020-05-25
SRE: Parse the Google Meet JWT from env var string.

cmfcruz:
The env var is considered as a string so we need to parse it into a JSON object.

Change-type: patch
Signed-off-by: Carlo Miguel Cruz <carloc@balena.io>

***

test suite!


grahammcculloch (commented):
@cmfcruz we are actually already calling `JSON.parse` - see [this line](https://github.com/product-os/jellyfish/blob/master/lib/action-library/handlers/action-google-meet.js#L25). I've verified it works correctly on my environment with the env var set as a string as discussed in flowdock.

PR #3860, opened 2020-05-25
ui: Only create a notification for 'insert' stream updates

grahammcculloch:
Change-type: patch

***

Relates to: #3789 (note: this PR doesn't fix the issue - but does reduce the problem!)

Previously we were calling the 'create notification' code on every type of update to a message/whisper card. When a message is sent, the stream listener is notified about an `insert`  event and then at least one `update` event. We only care about the initial `insert` event.


LucianBuzzo (approved):
nice!

ljewalsh (approved):

PR #3898, opened 2020-05-28
Add new group card

grahammcculloch:
Change-type: minor

***

Relates to: #2883

This card simply represents a group. In subsequent PRs we will associate users with groups and add support for pinging a group in order to notify everyone in that group.


LucianBuzzo (changes requested):
Can you please add the card as a JSON file here https://github.com/product-os/jellyfish/tree/master/apps/server/default-cards
and load the card using this file https://github.com/product-os/jellyfish/blob/master/apps/server/card-loader.js
The cards in `/lib/core` should only include the cards that are absolutely necessary for the Jellyfish kernel to run

LucianBuzzo (approved):

PR #3936, opened 2020-05-29
Improvements to org management

LucianBuzzo:
This PR includes a number of improvements and bug fixes:

- Generalise the CardChatSummary component to work with more card types, instead of being specific to support threads
- When viewing related cards ensure that add and link buttons are always visible
- Refactor create user form, to let you choose the org and to be less brittle

image


StefKors (dismissed):

grahammcculloch (approved):
Re-approved after toggle of pr-draft tag

PR #3937, opened 2020-05-29
Display assign to self if card is unassigned

karaxuna:
Change-type: minor
Connects-to: #3910

***

test suite!


grahammcculloch (approved):
LGTM!

Issue #3908, opened 2020-05-28
If a teardown summary already exists, use it to populate the teardown flow


LucianBuzzo:
When going through the support thread teardown flow after closing it, if there are already summary fields on the card, they should be used to auto populate the form.

PR #3951, opened 2020-06-01
Add summary event card

karaxuna:
Change-type: minor
Connects-to: #3907, #3908

***

**NOTE!** Marking summaries as read does not work. Depends on: https://github.com/product-os/jellyfish-client-sdk/pull/19


grahammcculloch (approved):

ljewalsh (changes requested):
We talked about the summary workflow at an architecture call a few weeks ago and one thing that was brought up was that it might be better to store the summary as a separate card so that it can become part of the event flow of a thread, rather than storing it on the thread itself. This seems to better represent the way we are thinking about card updates as a series of events in the Jellyfish system

LucianBuzzo (changes requested):

LucianBuzzo (approved):

ljewalsh (approved):

PR #3963, opened 2020-06-02
Add options for triggered actions to be sync or async

LucianBuzzo:
This change allows for triggered-actions to be evaluated and executed
asynchronously, resulting in a big performance boost when processing
actions. Triggered action cards default to being asynchronous.

Connects-to: #3697
Change-type: minor

***

test suite!


StefKors (approved):

StefKors (commented):
@LucianBuzzo The test fails on slug mismatches I think these should be reverted?

PR #3972, opened 2020-06-03
Update getUser trigger to find users by first and last name of a user as well as slug

ljewalsh:
At the moment when you type @ in a jellyfish thread, it triggers a search to find existing users by slug. This PR updates the trigger so that we also search for users by their name (first or last) if present.

I have also updated the component the trigger renders so that the full name of the user appears alongside the tag when a name is present.

https://www.loom.com/share/1e18468e1cfe4b128d89748d9df58e98

**NOTE:** A bunch of refactoring has been done in this PR to remove duplicate code and make the triggers simpler for testing. I have kept this as a separate commit so that it's easy to track the differences


grahammcculloch (changes requested):
Sorry @ljewalsh I spotted one more small thing to fix up on the query. 

Apart from that its good to go!

grahammcculloch (changes requested):

grahammcculloch (changes requested):

LucianBuzzo (approved):

ghost (approved):

grahammcculloch (approved):

PR #3976, opened 2020-06-03
Add support for skipping, limiting, and sorting linked cards in a query

Ereski:
This commit adds support for an additional key `links` in the `options` argument for `Kernel.query()` and `jsonschema2sql`. `links` is a map from link type to an `options` object for that link type. For example, passing the following as `options`:

```
{
  skip: 20,
  limit: 10,
  sortBy: 'updated_at',
  sortDir: 'desc',
  links: {
    'has attachment': {
        skip: 100,
        limit: 500,
        sortBy: 'created_at',
        sortDir: 'desc',
    }
  }
}
```

Root `skip`, `limit`, `sortBy`, and `sortDir` are the usual accepted keys for `options`, while the same keys inside `links['has attachment']` will only apply to the `has attachment` linked cards.

This is implemented through windowing functions over a list of linked card IDs, so overhead should not be significant.

Closes #3949

Change-type: minor

***

test suite!


LucianBuzzo (approved):
Some nitpicks, otherwise LGTM :+1:

Issue #3736, opened 2020-05-12
Graphile errors are polluting queue-related test logs in balena-ci


ljewalsh:
<img width="1229" alt="Screen Shot 2020-05-12 at 12 12 07 PM" src="https://user-images.githubusercontent.com/6325359/81624284-e55fb680-9449-11ea-9c8b-d8d19737b822.png">

PR #3983, opened 2020-06-04
ci: Update queue integration tests to use a shared backend

joshbwlng:
Change-type: patch

***

This PR:
- modifies queue integration tests to use a shared backend instance instead of creating a new one for each test
- moves slug/id generation utils to a shared location

These changes will help clear up the test log noise described in https://github.com/product-os/jellyfish/issues/3736 and will speed up tests a bit. Going to be opening PRs for other integration tests as well.

### DIFF
--- a/test/integration/core/helpers.js
+++ b/test/integration/core/helpers.js
@@ -4,18 +4,9 @@
  * Proprietary and confidential.
  */
 
-const uuid = require('uuid/v4')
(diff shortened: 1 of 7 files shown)


LucianBuzzo (dismissed):
Nice! Any idea how much faster they run?

LucianBuzzo (approved):

PR #3995, opened 2020-06-05
Update mentions, alerts and tags when editing messsages

grahammcculloch:
Change-type: patch

***

If you edit a message/whisper, you might modify the tags/mentions/alerts. These fields are now re-calculated and updated on the card when saving an edited message.


StefKors (changes requested):
@grahammcculloch Let's add a test for updating mentions alerts and tags when editing messages. It seems like something that will break/be forgotten when refactoring code like this.

LucianBuzzo (approved):

StefKors (approved):

PR #3998, opened 2020-06-05
Overlay favicon with mentions count

grahammcculloch:
Change-type: minor

***

Closes #3975

This nifty little component uses `react-helmet` to update the page's favicon link with a dynamically created image (as base64 data) that overlays the jellyfish icon with the count of mentions (unread messages).

The bulk of the cleverness is done in a reusable React hook, `useLabeledImage`, that is responsible for drawing the label (mentions count) on top of the base image on a HTML5 canvas.

If there are no unread messages, no overlay is added (i.e. we do not display `0` like Gmail does). If there are more than 99 unread messages, `99+` is displayed.

Check it out in operation (look at the icon in the tab header):
Favicon Count

### DIFF
--- a/apps/ui/JellyfishUI.jsx
+++ b/apps/ui/JellyfishUI.jsx
@@ -32,6 +32,7 @@ import Notifications from './components/HOC/Notifications'
 import RouteHandler from './components/RouteHandler'
 import Oauth from './components/HOC/Oauth'
 import Login from './components/HOC/Login'
+import CountFavicon from './components/CountFavicon'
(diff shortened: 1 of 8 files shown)


StefKors (commented):
Having dealt with canvas before, I love how clean `apps/ui/hooks/use-labeled-image.js` is. 👍 

I've left a Nit on `apps/ui/components/CountFavicon/index.js` let me know what you think but happy to approve it either way.

StefKors (approved):

PR #4015, opened 2020-06-08
backend: Fix queue saturation metrics call placement

joshbwlng:
Change-type: patch

***

Closes https://github.com/product-os/jellyfish/issues/3999

This PR fixes the placement of the queue job enqueue/dequeue/duration metrics calls. The previous location caught some job data, but was not deep enough to catch __all__ queue job data. These changes place metrics hooks inside of `lib/queue` to catch when jobs are added to the queue, when job results are returned, and how long it took for the job to get through the queue.


ljewalsh (approved):
LGTM!

PR #4034, opened 2020-06-09
backend: Move queue metric calls

joshbwlng:
Change-type: patch

***

Move queue saturation and duration metric calls to inside the graphile runner task list handler to make sure we don't miss any job increments or decrements.

There are cases, such as webhook requests, in which `waitResults` isn't called. This means that the metric would be incremented, but not decremented with the current code.

Was able to confirm that all worker integration tests trigger metric calls, translate tests also correctly trigger both increment and decrement metric calls. And common actions such as create-card, update-card, create-session, create-event are properly incremented and decremented.

queue-metrics

### DIFF
--- a/lib/queue/consumer.js
+++ b/lib/queue/consumer.js
@@ -11,6 +11,7 @@ const graphileWorker = require('graphile-worker')
 const logger = require('../logger').getLogger(__filename)
 const CARDS = require('./cards')
 const environment = require('../environment')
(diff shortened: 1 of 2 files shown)


grahammcculloch (dismissed):

grahammcculloch (approved):
LGTM. Out of interest, how did you hook up Grafana to your local environment to generate that screenshot in the PR description?

PR #4036, opened 2020-06-09
MessageInput: Allow pasting files

karaxuna:
Change-type: minor

***

test suite!

### DIFF
--- a/lib/ui-components/Timeline/MessageInput.jsx
+++ b/lib/ui-components/Timeline/MessageInput.jsx
@@ -107,6 +107,15 @@ const MessageInput = React.memo(({
 }) => {
 	const theme = useTheme()
 
+	const handlePaste = React.useCallback((event) => {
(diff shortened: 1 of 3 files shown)


StefKors (changes requested):
Could you add a unit test for this?

grahammcculloch (commented):
I like how simple this change is! 

However, I did test this out locally and I'm a bit worried about the behaviour. It's super easy to accidentally upload an image from the clipboard (e.g. when you thought you had some text in the clipboard but it's actually an image, or you just accidentally hit `Ctrl+V`). And once it's uploaded, there's (currently) no way to delete that message (via the UI). I would much rather see a solution that uploads an image and creates a markdown image reference to it in the message body - just like GitHub does. I guess that is more tricky in terms of authentication when fetching the uploaded image. 

But maybe this is an acceptable intermediate solution?

grahammcculloch (approved):

StefKors (approved):

PR #4048, opened 2020-06-10
Add index for endpoint field of a web-push-subscription

grahammcculloch:
We will want to query web-push-subscriptions by endpoint when checking for existing subscriptions.

Change-type: patch

***

When we start-up the front-end we fetch the existing push subscription from the `PushManager`. If there is a subscription we check that we've cached the subscription on the server by querying for a `web-push-subscription` card with that endpoint specified that is also linked to the authenticated user. For this reason we want to index the `data.endpoint` field of a `web-push-subscription` card.

### DIFF
--- a/apps/server/default-cards/contrib/web-push-subscription.json
+++ b/apps/server/default-cards/contrib/web-push-subscription.json
@@ -34,6 +34,9 @@
           }
         }
       }
-    }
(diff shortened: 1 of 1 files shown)


karaxuna (approved):
LGTM

PR #4081, opened 2020-06-11
Livechat: Fix actor click navigating to blank page

karaxuna:
Change-type: #patch
Connects-to: #4057

***

See: https://jel.ly.fish/79075819-1c05-4500-8afb-18d1c4450b9b

### DIFF
--- a/apps/ui/components/Inbox/MessageList.jsx
+++ b/apps/ui/components/Inbox/MessageList.jsx
@@ -13,7 +13,8 @@ import {
 	withRouter
 } from 'react-router-dom'
 import {
-	bindActionCreators
+	bindActionCreators,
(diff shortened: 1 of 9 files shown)


grahammcculloch (dismissed):
Left one comment about not wrapping actor's name in a `Link` if we're not actually going to link to anything.

Also, I think the following components reference `Event` and will need to pass in the `getActorHref` prop with a valid callback method:
* `apps/ui/components/Inbox/MessageList.jsx`
* `apps/ui/lens/full/SupportThread/index.jsx`
* `apps/ui/lens/list/Interleaved.jsx`

LucianBuzzo (approved):

PR #4089, opened 2020-06-12
Constrain size of images in Jellyfish threads

ljewalsh:
At the moment, images are huge in Jellyfish. This PR enables a `maxImageSIze` prop to be passed down to the `AuthenticatedImage` component. This is used to set a `maxWidth`  and `maxHeight` on the image. `maWidth` is set using `min(maxImageSize, 100%)` so that it falls back to 100% when the `maxImageSize` is too wide for the channel

This `maxWidth` and `maxHeight` have also been set in the same way on the`Markdown` images so that they also constrain to this size.

**NOTE**: I took the opportunately to move the attachment-related code from the `EventBody` component into its own component called `Attachments`. I have moved some related tests here too


grahammcculloch (changes requested):
I noticed Lucian's comment about leaving on-click behaviour until we can be consistent for all images. Sounds sensible.

Just one request for a change regarding setting `max-height` as well.

grahammcculloch (commented):
Looking good now @ljewalsh 👍 . If you update the PR title and description to reflect the changes and then squash your commits as you see fit I'll go ahead and approve this.

grahammcculloch (approved):

PR #4092, opened 2020-06-12
Concatenate messages from the same user on the same thread

LucianBuzzo:
Fixes #3964

This change helps streamline the UI when there are multiple consecutive messages:
**Before**
image

**After**
image

Change-type: minor

***

test suite!


joshbwlng (approved):
Tested locally with Chrome and Firefox, LGTM

grahammcculloch (changes requested):
The major problem I see here is that you now can't edit a message (or copy it as JSON/raw text) etc if it follows another message you have written. Or am I missing something?

grahammcculloch (approved):

PR #4096, opened 2020-06-14
backend: Handle init db duplicate errors

joshbwlng:
Change-type: patch

***

This PR:
- adds logic to check for and allow specific errors to occur during initial db setup
- adds retry logic when starting the Graphile worker

in order to handle the errors described in https://github.com/product-os/jellyfish/issues/4095: catch and handle Postgres `duplicate key` and `unique violation` errors that occur when multiple services are starting up at the same time and attempting to execute the same SQL simultaneously

Closes https://github.com/product-os/jellyfish/issues/4095

### DIFF
--- a/lib/core/backend/postgres/index.js
+++ b/lib/core/backend/postgres/index.js
@@ -387,8 +387,14 @@ module.exports = class PostgresBackend {
 			 * it again, but then we might fail to do the operations we need
 			 * because it doesn't belong to us.
 			 */
-			await this.connection.any(`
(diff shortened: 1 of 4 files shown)


grahammcculloch (changes requested):
This could make the CI tests much more reliable! 

Couple other notes:
1. It would be great to abstract a generic 'retry' wrapper method as we use this pattern a number of times throughout our code-base. I'm surprised lodash doesn't have one (I checked but couldn't find it).
2. Why do we not use `IF NOT EXISTS` on those SQL statements that can clash?

grahammcculloch (approved):

PR #4097, opened 2020-06-14
Move Prometheus and Grafana into separate compose config

joshbwlng:
Change-type: patch

***

This PR:
- separates the Prometheus and Grafana services to a separate Docker Compose config
- adds a note about how to start Prometheus and Grafana to README

The problem with including Prometheus and Grafana in the main Docker Compose config is that this config is used when running tests in balenaCI. As Prometheus and Grafana are not necessary to run Jellyfish or its tests, moving these services to a separate file will speed up CI a bit.

To run Jellyfish with Prometheus and Grafana, just set `MONITOR=1` and the new `docker-compose.monitor.yml` will be included in the list of files to use.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -249,8 +249,13 @@ export CI
 VISUAL ?=
 export VISUAL
 
+DOCKER_COMPOSE_FILES = --file docker-compose.yml
+ifdef MONITOR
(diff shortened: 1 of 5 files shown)


jimsynz (approved):
LGTM

PR #4101, opened 2020-06-15
ui: Cache groups as part of bootstrap actions

grahammcculloch:
The groups are saved in the redux state, with the list of group users and a boolean flag indicating whether the authenticated user is part of that group (for efficient querying of groups that the user is part of).

Change-type: minor

***

Relates to: #2883 

In preparation for i) formatting group mentions in messages based on whether the authenticated user is in the group or not and ii) pinging the authenticated user when a message is sent that mentions a group they are part of, the front-end now caches group cards (with member users) on bootstrap. This will enable the front-end to efficiently do i) and ii) (to be implemented in subsequent PRs) without querying the backend each time.

A new unit test in `reducer.spec.jsx` shows an example of what [the cached redux groups data looks like](https://github.com/product-os/jellyfish/blob/49bdfb7ae2e8dec92685cb62ae4cd5e5792d2965/apps/ui/core/store/reducer.spec.js#L644-L654) 

_Note: as groups and group-memberships are added/modified, we could always listen for streamed updates and update the cache based on this. But as group membership is fairly static it's a reasonable start to just cache these details when the front-end is loaded._


LucianBuzzo (approved):
LGTM :+1:

PR #4107, opened 2020-06-16
When a synced support thread is created, post a whisper containg a link to JF

LucianBuzzo:
This change makes it easier to find the corresponding Jellyfish thread
when viewing an external resource like Front, Discourse or Flowdock

image

Change-type: minor

***

test suite!


StefKors (commented):
@LucianBuzzo Is it necessary to have a test for this?

StefKors (approved):

PR #4195, opened 2020-06-22
Use queries to paginate timeline

ljewalsh:
When we load a single card at the moment, we query the entire timeline and then show slices of it using frontend pagination. This PR moves to paginating the timeline through queries rather than the UI.

**Basic behaviour**
When the single page is loaded, we query for the card alongside 20 of the latest items on the timeline. These items are displayed when the timeline loads. When the user scrolls upwards, a query is triggered for 20 more items until we reach the beginning of the timeline and render a "Timeline Beginning" message. I used the `InfiniteList` component to handle the scrolling triggers and updated it so that it can be used in reverse.

**Additional behaviour:**
A few other things have had to be updated as part of this behaviour change:

1. Most substantially, I have updated `InfiniteScroll` so that it can continue to query for data when the scroll area has not overflowed. This is necessary because of how the toggles work in the timeline. For example, in a case where you have 18 whispers and 2 messages in your first list of results, if you toggle the whispers off, then suddenly you only have two items on your timeline. This is not enough items for the container to overflow and so the user cannot scroll upwards to trigger more queries. From the user's perspective, it looks like they have reached the start of the timeline when we have actually got stuck at the end of the timeline (there is quite possibly more messages in the past but we can't query for them).

I toyed with the idea of making the container always scrollable but it was difficult to make a solution like this that wasn't unwieldy. In the end, it seemed more sensible to continue querying for more data until we have reached either the start of the timeline or we have filled the timeline container.

2. The handleJumpToTop now runs a query to retrieve all the data before jumping to the top

3. I implemented some code for scrolling to specific events (in the instance that the user puts a direct link into the url). First I checked that the event is present in our page of results. If it is not, I retrieve all the data and then scroll to the event. This may not be the best solution but I thought it was potentially better to do it this way than to load page after page of results for an event that could be right at the beginning of the timeline

4. I cleaned up how we are handling pending messages (specifically how we remove them from the state once they have been processed)

### DIFF
--- a/apps/ui/core/store/actioncreators/index.js
+++ b/apps/ui/core/store/actioncreators/index.js
@@ -526,7 +526,23 @@ export default class ActionCreator {
 				}
 
 				return this.sdk.card.getWithTimeline(target, {
-					type: cardType
+					schema: {
+						type: 'object',
+						properties: {
+							type: {
+								const: cardType
+							}
+						}
+					},
+					queryOptions: {
+						links: {
+							'has attached element': {
+								limit: 20,
+								sortBy: 'created_at',
+								sortDir: 'desc'
+							}
+						}
+					}
 				})
 					.then((result) => {
 						if (!result) {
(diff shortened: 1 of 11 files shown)


StefKors (approved):
LGTM :shipit:

PR #4212, opened 2020-06-23
Randomize metrics port number on collision

joshbwlng:
Change-type: patch

***

This PR:
- randomizes the app metrics port to a random one when the default port is already in use
- confirms that `METRICS_PORT` and `SOCKET_METRICS_PORT` are numbers


karaxuna (commented):
I would not assign a default value if `METRICS_PORT` does not exist, instead I would directly run metrics server on random port if env var is empty. But probably this is not really important.

ghost (approved):

PR #4240, opened 2020-06-25
Livechat: Display commands sent by support agents

karaxuna:
Change-type: minor

***

test suite!


joshbwlng:
@karaxuna Can we delete this? It hasn't been updated in over 2 years.

PR #4241, opened 2020-06-25
Switch to third-party library for UUID validation

joshbwlng:
Change-type: patch

***

This PR switches to using [`is-uuid`](https://www.npmjs.com/package/is-uuid) for UUID validation. 
This is a part of a series of PRs to break apart our monorepo.


grahammcculloch (approved):
This change looks fine to me.

Of course - it will be fairly moot if we don't also manage to remove the other exports from `lib/uuid`... What's your plan for replacing the async `uuid.random()`?

PR #4243, opened 2020-06-25
ci: Move sidecar to separate compose config

joshbwlng:
Change-type: patch

***

Move `sidecar` to a separate compose config that is used only when `SIDECAR=1` is set.

`sidecar` is only used in CircleCI and therefore wastes precious time and resources when unnecessarily built and ran in balenaCI.
We are also planning on doing away with `sidecar` once all tests are moved to balenaCI, so this will make that change quite easy.


ljewalsh (approved):
LGTM

PR #4255, opened 2020-06-26
backend: Add optional options parameter to server bootstrap

joshbwlng:
Change-type: patch

***

This allows for setting custom options, specifically backend constructor options, when setting up a server instance.

Normally environment variables will suffice, but there are times, especially with tests, in which we need to customize parameters on a case-by-case basis in the code.


grahammcculloch (changes requested):
One small fix suggested.

grahammcculloch (approved):

PR #4270, opened 2020-06-29
Ensure group names contain only certain characters

grahammcculloch:
We want to ensure that group names don't contain any spaces or other odd characters. 

Change-type: minor

***

Note: We don't currently have any groups defined so it shouldn't cause any problems to add this restriction.

Ideally we also want to make group names unique. I can't see a way to do this at the moment with JSON schema?


ljewalsh (dismissed):
@grahammcculloch can we add a test for this?

LucianBuzzo (approved):

PR #4297, opened 2020-07-01
Replace lib/assert with private package

joshbwlng:
Change-type: patch

***

Require [`@balena/jellyfish-assert`](https://github.com/product-os/jellyfish-assert) instead of `lib/assert`.
This is a part of breaking up the effort to break up the monorepo.


grahammcculloch (changes requested):
Think you need to update `ARCHITECTURE.md` and `package.json` to remove references to lib/assert? Other than that, looks good.

grahammcculloch (approved):

PR #4316, opened 2020-07-02
Refactor timeline to make it a little easier to work with

ljewalsh:
Refactor of the Timeline component done in order to make pagination a little easier. Reviewing by commit should be the easiest approach.

Work includes:
- Moving header-related code into its own component (`Header`) - this could be broken-down further but I've left this for future work
- Moving pending-message-related code into its own component (`PendingMessages`)
- Moving userTyping code into its own component (`TypingNotice`)
- Refactoring how we figure out which events to display. Previously we were looping through the list to remove the updates/create events and then doing another loop to check to see if the whisper should be removed. I've moved all the event code into a component called `EventList` which does one loop and makes decisions about how to render the card within that loop. It also compartmentalises this code into one place. I've added some tests to make sure this is correct.

Also closes #4071

### DIFF
--- a/lib/ui-components/Timeline/EventsList.jsx
+++ b/lib/ui-components/Timeline/EventsList.jsx
@@ -1,0 +1,92 @@
+/*
+ * Copyright (C) Balena.io - All Rights Reserved
+ * Unauthorized copying of this file, via any medium is strictly prohibited.
+ * Proprietary and confidential.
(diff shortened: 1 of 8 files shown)


grahammcculloch (approved):
Ahhh - SO much cleaner now 😍 

Great work!

PR #4332, opened 2020-07-06
test: wait for thread sync whisper after creating support-thread

grahammcculloch:
This should fix the race condition seen where sometimes the first 'test' message in a thread is saved before the 'sync whisper' that is created by the asynchronous triggered action. This makes it hard to determine which event will be the 'last' in the list of events. Hopefully by waiting for the sync whisper before posting any test message to the thread, we ensure the sync message will always be the first one in the thread.

Change-type: patch

***

An example of the type of test failure this PR should fix can be found **[here](https://app.circleci.com/pipelines/github/product-os/jellyfish/44842/workflows/cf57e1f1-7ef2-4011-a08d-213aeecc678d/jobs/135858/steps)**

### DIFF
--- a/apps/server/default-cards/contrib/triggered-action-sync-thread-post-link-whisper.json
+++ b/apps/server/default-cards/contrib/triggered-action-sync-thread-post-link-whisper.json
@@ -65,11 +65,10 @@
       "$eval": "source.links['is attached to'][0].id"
     },
     "arguments": {
-      "reason": null,
       "payload": {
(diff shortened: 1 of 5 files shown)


jimsynz (dismissed):
I'm in

joshbwlng (changes requested):
@grahammcculloch Looks like something didn't work: https://circleci.com/gh/product-os/jellyfish/135953

joshbwlng (approved):

PR #4335, opened 2020-07-06
Replace lib/environment with private package

joshbwlng:
Change-type: patch

***

Replace `lib/environment` with private package [`@balena/jellyfish-environment`](https://www.npmjs.com/package/@balena/jellyfish-environment).
Part of the work to break up the monorepo.


grahammcculloch (changes requested):
Think you just need to update the Architecture.md file to remove the reference to `lib/environment`. Other than that, looks fine :-)

grahammcculloch (approved):

PR #4348, opened 2020-07-07
Switch to using a private package for lib/logger

joshbwlng:
Change-type: patch

***

Replace `lib/logger` with private package [`@balena/jellyfish-logger`](https://github.com/product-os/jellyfish-logger)

The move to a separate package means that we can no longer get the Jellyfish version from `package.json`. Many `context.id` strings already contained this version information so I went ahead and added it to other `context.id` strings in this PR and removed the `packageJSON.version` output from logs [here](https://github.com/product-os/jellyfish-logger/pull/4).


grahammcculloch (commented):
It seems weird to include the app version in the context `id` string. Can we not include it as a separate context `version` variable? And strip it out of any existing id strings? 

That may be a lot of tedious refactoring in which case I'm fine with making that a separate task/PR. But just wanted to check why we're using the version in the `id` field.

StefKors (approved):

grahammcculloch (approved):
@joshbwlng I'll approve this change now but I think we should probably open an issue to fix the log context by using separate fields for app version (and app name).

PR #4416, opened 2020-07-14
Add relationship type-cards for all known relationships.

jimsynz:
This PR adds a bunch of new "relationship" typecards, which are based on the contents of [the SDK's link constraints file](https://github.com/product-os/jellyfish-client-sdk/blob/master/lib/link-constraints.js). This is work that is needed to make the GraphQL link schemas fully generatable.

So, this PR includes a few important things:

* A whole bunch of new typecards which define the valid relationships between specific types of cards.
* A new `linkCards` method on kernel which handles the creation of links automatically, rather than generating link cards manually.
* A deprecation warning on `insertCard` if the inserted object is of type `link` including a stack-track to aid in identifying all the places where we create link cards.
* An MVP migrations system with a migration written to which matches and updates the `type` property on existing link cards to the correct type.

## New relationship cards:

You'll notice a bunch of new cards in `app/server/default-cards/relationships`.  You'll also notice that they make use of @grahammcculloch's fancy pluggable loading system to dynamically generate the card's JSON Schema based on the other values in the card's data.

### Relationship data:

- `is_relationship` this is currently used to identify that not only is this a type card, but a special relationship card. This will be used by GraphQL to identify these cards for type generation.
- `forward` the name of the association in the forward direction (eg "X is owner of Y").
- `reverse` the name of the association in the reverse direction (eg "Y is owned by X").
- `type_pairs` is a list containing tuples of types in the form of two-element arrays. Pairs are specified in the forward direction (eg `[ 'X', 'Y' ]` in the examples above). Each tuple element can either be the (unversioned) name of a type as a string (slug form) or an object containing a `name` and `title` property for cases where inflecting the association name from the type name wouldn't be appropriate (eg `['group', {name: 'user', title: 'Member'}]`).

### Relationship schema:

The relationship schema is generated based on the `forward`, `reverse` and `type_pairs` properties and generates a JSON schema which matches cards which link any cards of the specified types.

For example, the following relationship:

```
{
  forward: 'is of',
  reverse: 'has',
  type_pairs: [
    ['checkin', 'project'],
    ['thread', 'repository]
  ]
}
```

Would match any of the following link cards:

1. `checkin is of project`
2. `project has checkin`
3. `thread is of repository`
4. `repository has thread`

In this way a single relationship can represent links in both directions.  The `linkCards` kernel method takes this direction information into account when creating link cards.

## The `linkCards` method

This method is primarily just a DX improvement, but should hopefully help us eliminate places where non-complying links are generated in the system.  When provided two cards and the relationship type (eg `relationship-is-observed-by-observes@1.0.0`) it creates a link ~~in the direction matching the order which the cards were passed as arguments~~ with the direction normalised as per the relationship card.  Because this requires a search of the relationship's `type_pairs` array, we can also short-circuit type errors without having to rely on JSON-Schema to validate the entire link card.

## Deprecation warning on `insertCard`

When `insertCard` detects that an old-style link card is being inserted it logs a warning which includes a backtrace to aid in tracking down places in our code which manually create link cards.  At least on the backend this is overwhelmingly in tests, so while I expect it to happen in production quite a lot initially, they should be easy enough to clean up.

## Migrations

Database migrations should probably be a solved problem, but since we don't seem to be able to form a consensus about how to handle it I have implemented a migration for our existing link cards both because we'll need to run this in production at some stage before we will be able to see any historical data via the GraphQL API and because it's easier to have a discussion about existing code than imagined code.

## After this lands

This is just part of an on-going work stream, so there's a lot to follow on from it;
- Modify the client SDK to use these relationship types to generate the link constraints table rather than using a hard-coded one.
- Modify the GraphQL type generation system to generate all necessary type objects which can map from each type to each of it's connected nodes.
- Continue hunting down and updating any code which triggers our deprecation warning.

### DIFF
--- a/apps/server/bootstrap.js
+++ b/apps/server/bootstrap.js
@@ -138,23 +138,11 @@ module.exports = async (context, options) => {
 			jellyfish.errors.JellyfishNoElement,
 			`Test org does not exist: ${environment.test.user.organization}`)
 
-		await jellyfish.replaceCard(
-			context, jellyfish.sessions.admin, {
-				type: 'link@1.0.0',
-				name: 'has member',
-				slug: `link-${orgCard.id}-has-member-${userCard.id}`,
-				data: {
-					inverseName: 'is member of',
-					from: {
-						id: orgCard.id,
-						type: orgCard.type
-					},
-					to: {
-						id: userCard.id,
-						type: userCard.type
-					}
-				}
-			})
+		await jellyfish.linkCards(context,
+			jellyfish.sessions.admin,
+			orgCard,
+			userCard,
+			'relationship-is-member-of-has-member@1.0.0')
 	}
 
 	logger.info(context, 'Configuring socket server')
--- a/apps/server/default-cards/index.js
+++ b/apps/server/default-cards/index.js
@@ -145,10 +145,11 @@ module.exports = {
 		require('./relationships/relationship-is-backup-owner-of-has-backup-owner')(mixins),
 	relationshipIsContributedToByContribuesTo:
 		require('./relationships/relationship-is-contributed-to-by-contributes-to')(mixins),
+	relationshipIsExecutedByExecutes: require('./relationships/relationship-is-executed-by-executes')(mixins),
 	relationshipIsFeedbackForIsReviewedWith: require('./relationships/relationship-is-feedback-for-is-reviewed-with')(mixins),
 	relationshipIsGroupMemberOfHasMember: require('./relationships/relationship-is-group-member-of-has-member')(mixins),
 	relationshipIsGuidedByGuides: require('./relationships/relationship-is-guided-by-guides')(mixins),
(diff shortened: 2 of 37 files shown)


grahammcculloch (changes requested):
Nothing major to change. I think the main query I had was about throttling the migration script.

I did have one suggestion about this change though... 

We're adding a whole load of new type cards. The frontend on bootstrap [fetches _all_ type cards](https://github.com/product-os/jellyfish/blob/261f49b9f9aeab3a47c824f703f4bcea17d51df7/apps/ui/core/store/actioncreators/index.js#L777). But currently won't need to cache these new relationship type cards. To avoid unnecessarily fetching all these relationship type cards we should probably fix this behaviour so we filter on type cards where `data.isLink` is not `true`. This will ensure the frontend startup isn't slowed down too much and the cache of type cards is kept 'lean'.

PR #4417, opened 2020-07-15
Remove /metrics endpoint test

joshbwlng:
Change-type: patch

***

Moving the `/metrics` endpoint test to `product-os/jellyfish-metrics` with https://github.com/product-os/jellyfish-metrics/pull/11.


grahammcculloch (changes requested):
Do you also need to remove the `test/integration/server/metrics.spec.js` file? I realise it tests for socket _and_ app metrics so may be different.

grahammcculloch (approved):

PR #4419, opened 2020-07-15
Define pattern card type

karaxuna:
See: https://trello.com/c/7wdpohSc

test suite!


grahammcculloch (commented):
@karaxuna will the `pattern` type be able to have events (e.g. a timeline)? If so, I would suggest using the same approach as the `issue` type card (for example) which makes use of the `withEvents` mixin.

grahammcculloch (approved):

PR #4425, opened 2020-07-15
Form response card view

mbalamat:
Screenshot 2020-07-17 at 4 35 17 PM

test suite!


grahammcculloch (changes requested):
Should be easy enough to make these changes :-)

grahammcculloch (changes requested):
@mbalamat can you add an e2e test for these lenses? You can copy the approach used in `test/e2e/ui/support.spec.js`. Give me a shout if you'd like to pair-program on this.

Just one nit-pick on the UI code to fix (see comment)

grahammcculloch (changes requested):
Beautiful test - love it! 

One nit-pick: can you fix the indentation on the `users-form-response.json` fixture file to be 2 spaces rather than 4 and then move it from the `assets` folder to a new sibling `fixtures` folder? Just to keep things consistent :-)

grahammcculloch (approved):
LGTM!

PR #4434, opened 2020-07-16
core: simplify the streaming subsystem by relying on the common `query()` codepath

Ereski:
This also removes the `before` key on stream events.

Change-type: major


LucianBuzzo (changes requested):

grahammcculloch (commented):
Is there any way to break this PR into smaller chunks? Or are all these UI changes needed in the same release as the backend changes? This is just a terrifyingly large PR to digest :-)

LucianBuzzo (approved):

PR #4438, opened 2020-07-17
Add more field to settings/profile

juanfridano:
Change-type: patch
Signed-off-by: Juan Fernandez Ridano <juanfr@balena.io>

***

test suite!


StefKors (approved):
LGTM :shipit:

PR #4443, opened 2020-07-20
ui: Ensure 1-to-1 messages appear in the user's inbox

grahammcculloch:
And also in the mention count!

Change-type: minor

***

Closes #4361

Note - this PR might result in people having a LOT of unread messages if they have a lot of 1-to-1 conversations - as these were previously not marked as read. But just going into your inbox and clicking on the 'Mark all read' button will get you back to normal.


ljewalsh (approved):
This looks good to me. Tested it locally too

Issue #3455, opened 2020-04-02
Sync orgs from balenaAPI


StefKors:
| Balena API entity | Jellyfish Card |
| ----------------- | -------------- |
| Orgs              | Accounts       |

It should be a 1-way mirror from the Balena API to Jellyfish. If multiple users are part of an org they should be linked together. In Jellyfish possibly extra information and linked data will be added.

- Omit billing data

Questions:
- **How to sync account import with account card schema?**

In the future, the google appscript that syncs information from recurly to google sheets will be ported to jellyfish. 

Meeting notes and links:
- https://www.flowdock.com/app/rulemotion/p-billing/threads/fL8CTiNnTKOgFw7UBI8tIMD1Wkh
- https://docs.google.com/document/d/1VGyWMf6YDNs_6gLnuJB3AA0MzMzMMUe_7Kk98B04VYM/edit
- https://docs.google.com/document/d/1VGyWMf6YDNs_6gLnuJB3AA0MzMzMMUe_7Kk98B04VYM/edit
- https://github.com/balena-io/balena-sdk/pull/621
![balena-api](https://user-images.githubusercontent.com/11800807/87927761-0402af00-ca84-11ea-9c21-4521cb84628b.png)

Billing Spec: https://github.com/balena-io/balena/pull/586/files

- [x] Setup dev env
- [x] Reactored the balena-api translate function
- [x] Add code to create accounts (Jellyfish)
- [x] Add code to send webhook from balena-api
- [x] FIX BALENA API EMPTY USER CREATE
- [ ] Link contact cards to account cards in Jellyfish
- [x] Check if updates sync correctly
       - When cards exist they updated, instead of creating new cards
       - Check if this updates the link between account cards and contact cards correctly

- [ ] Debrief on dev env setup (two NUCs)

## Extension goal
- [x] Expand the balena-api webhook to include billing data

The addons are Recurly subscriptions

PR #4460, opened 2020-07-20
Update balena-api integration to support organisations

StefKors:
Related-to: https://github.com/product-os/jellyfish/issues/3455
Change-type: minor

This adds support for translating balena-api organizations to sales accounts in Jellyfish. Before this works, it needs web-hooks in balena-api.


LucianBuzzo (changes requested):
Can you omit the `tradegecko_company_id` and `billing_account_code` fields from the synced data? We need to have a discussion about how we deal with the information these fields represent

LucianBuzzo (approved):

PR #4508, opened 2020-07-23
Add sort button to View lens

ljewalsh:
Adds a sort button on the view lens, allowing the user to sort the list of cards by fields on the card. Certain fields have been admitted since they do not make sense. These include: `id`, `type`, `version`, `active`, and any field that is an array.

I also took the time to refactor the `Table` lens to use the same helpers. These helpers are now at the top-level of the lens folder. Hopefully we can continue to use them for other lenses.

Still need to:

- [x] figure out why the sort doesn't occur until we scroll
- [x] ensure cards coming through the stream are sorted correctly


StefKors (approved):

grahammcculloch (dismissed):
@ljewalsh nicely structured 👍 

Found some mostly-React-specific things that probably need changing. Although I haven't tested this branch locally so I'm basing this on eye-balling the code only.

PR #4517, opened 2020-07-25
ui: Move some components apps/ui

joshbwlng:
Change-type: patch

***

Move a number of components from `lib/ui-components` to `apps/ui/components`.
Once this is merged, I can then move `lib/ui-components` and therefore `lib/chat-widget` out of the monorepo.

This list of what needs to be moved was provided by @grahammcculloch  in [this thread](https://jel.ly.fish/4abd208a-3465-447b-b326-aa13f74232e3?event=7e40c34e-17dc-4ff7-ac68-718fc332dfd1).


grahammcculloch (changes requested):
All looks good - just one change that wasn't obvious that should be made in order to keep things consistent. 

Instead of 
`apps/ui/components/HOC/Notifications.jsx` -> `apps/ui/components/Notifications.jsx`

can you crreate a new folder `apps/ui/components/Notifications` and move those two files into it as follows:
`apps/ui/components/HOC/Notifications.jsx` becomes `apps/ui/components/Notifications/index.jsx`
`apps/ui/components/Notifications.jsx` becomes `apps/ui/components/Notifications/Notifications.jsx`

You'll need to adjust imports to/from these files accordingly.

Same with `Oauth`. 

Once this is done you can actually delete the (now empty) `apps/ui/components/HOC` folder.

grahammcculloch (approved):

PR #4522, opened 2020-07-26
lens/actions/CreateLens: Order form response curation form

mbalamat:
Screenshot 2020-07-26 at 10 01 56 PM

With this, when creating a new form-response-curation, the card's [fieldOrder](https://github.com/product-os/jellyfish/blob/496a0415332641bd6919547a18e3e380e48b2556/apps/server/default-cards/contrib/form-response-curation.json#L71-L80) property is taken into account and the fields are rendered in the right order.

### DIFF
--- a/apps/ui/lens/actions/CreateLens.jsx
+++ b/apps/ui/lens/actions/CreateLens.jsx
@@ -381,6 +381,13 @@ class CreateLens extends React.Component {
 
 		const head = this.props.channel.data.head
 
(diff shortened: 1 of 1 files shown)


grahammcculloch (changes requested):
Actually @mbalamat I think the correct way to do this is to modify the `ui:order` property of the `uiSchema` object that is passed to the `Form` component. You shouldn't need to modify the `schema` at all. Should be much simpler/cleaner 😄

grahammcculloch (approved):
Nice! (Although this will hopefully be replaced soon when we add `uiSchema` to the card type)

PR #4565, opened 2020-08-03
Home channel performance improvements

grahammcculloch:
Change-type: patch
***

Various improvements to the rendering performance of the HomeChannel component and its sub-components. See individual commits for details.

Once this is reviewed I will squash those commits into one.


LucianBuzzo (approved):
Nice improvements! 👍

PR #4575, opened 2020-08-05
Replace lib/chat-widget with private package

joshbwlng:
Change-type: patch

***

Replace `lib/chat-widget` with private package `@balena/jellyfish-chat-widget`.


grahammcculloch (changes requested):
Just spotted one issue in the test/register.js file.

grahammcculloch (approved):

PR #4600, opened 2020-08-13
ci: Group failed e2e test screenshot comments

joshbwlng:
Change-type: patch

***

Group up failed E2E Livechat/UI test failure screenshot links into a single comment.

When we switch to using `ci-task-runner` to run our tests on balenaCI, we will be automatically retrying certain tasks when they fail. As the comments are created at the end of a test suite with ava's `after`, if `e2e-ui` or `e2e-livechat` tests are failing consistently, we will still get multiple comments. This change reduces the number of comments created per test suite run, limiting to a single comment per run.

See comments below for examples of what the comments look like. These were created by the code in this PR when failing certain tests on purpose.

### DIFF
--- a/test/e2e/livechat/helpers.js
+++ b/test/e2e/livechat/helpers.js
@@ -14,7 +14,15 @@ exports.afterEach = async ({
 	context, test
 }) => {
 	if (!test.passed) {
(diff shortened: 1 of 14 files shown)


grahammcculloch (approved):
Great stuff 👍

PR #4602, opened 2020-08-13
Add link to home from 404 view

grahammcculloch:
Link to home is only displayed if the channel is the only one (apart from the home channel). 

Refactored 404 view into it's own component.

Change-type: minor

***

Closes #4533 

TakeMeHome


karaxuna (approved):
LGTM :+1:

PR #4624, opened 2020-08-19
Fix pagination on the default list lens

StefKors:
Change-type: patch

***

test suite!


grahammcculloch (approved):
I'm trusting you've tested this locally Stef 👍

PR #4643, opened 2020-08-25
core: simplify the streaming subsystem by relying on the common `query()` codepath

LucianBuzzo:
***

test suite!


StefKors (approved):
beep boop 🤖  approving self certify for Lucian

PR #4649, opened 2020-08-26
Fix inbox pagination

ljewalsh:
I'm not sure if this is the best way of solving this. Essentially the `updatePage` captures the state of both `results` and `page` on first render. Even when these are updated, `updatePage` continues to use the stale values. To get around this, Carol has stored the results as a ref and we can access this. I could setup an additional ref for page but I felt it was easier to pass the old page through as an argument.

Great guide on stale closures here: https://dmitripavlutin.com/react-hooks-stale-closures/

### DIFF
--- a/apps/ui/components/Inbox/InboxTab.jsx
+++ b/apps/ui/components/Inbox/InboxTab.jsx
@@ -167,16 +167,15 @@ export default (props) => {
 				})
 			}
 
-			const updatePage = (newPage) => {
-				// If we haven't seen the maximum results, set the new page
+			const updatePage = (oldPage) => {
+				// If we have been returned the maximum results, set the new page
 				// TODO: Fix this hack once we can fetch a count from the API
-				if (options.limit * page === results.length) {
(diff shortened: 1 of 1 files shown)


grahammcculloch (commented):
Ok... so pagination on the inbox tabs isn't really pagination. It just fetches more and more results as you scroll down. For this to be pagination we need to keep `limit` fixed and just set `skip` based on the current page and the page size.

Otherwise we will overly load the server by repeatedly fetching more and more data as the user continues to scroll down!

I don't know if the new streaming query code has affected the whole area of pagination but in order for me to approve this PR I'd want to know it represents a very short-lived temporary hack solution to allow more results to be loaded before pagination is actually implemented (with an InfiniteList and react-virtualized).

(It's probably related but I'm also uncomfortable with ignoring the `page` argument of the `setPage` callback from `MessageList` and just assuming we want `oldPage + 1`. This needs to be replaced soon!)

Ereski (approved):

PR #4682, opened 2020-09-01
apps/server/default-cards/contrib: Add user-feedback card type

mbalamat:
This will replace the current form-response and form-response-curation.

Change-type: patch
Signed-off-by: Marios Balamatsias <mbalamatsias@gmail.com>

***

test suite!


grahammcculloch (commented):
Just left one question about this change - would love to hear the rationale!

grahammcculloch (approved):
LGTM

PR #4727, opened 2020-09-08
Ensure loadViewData and loadMoreViewData actions are shared between the inbox and the other views

ljewalsh:
**NOTE**: I would recommend reviewing this by commit rather than all at once

At the moment, the Inbox has its own bespoke code for loading and paginating through the messages. I have a suspicion that the notification bug is being caused by the lack of consideration for race conditions in the stream updates (the loadViewData action creates a queue to manage race conditions)

As part of a wider process to de-duplicate the loading and pagination of lists in Jellyfish, I decided to break the stream management out of loadViewData and loadMoreViewData into two new actions named `setupStream` and `paginateStream`. Extracting them out like this will help when we move stream pagination to the sdk.

This PR covers the extraction of these methods, as well as the work to update the `InboxTab` component to use the new methods. I also extracted out the `MarkAsReadButton` code to its own file

### DIFF
--- a/apps/ui/components/Inbox/Inbox.jsx
+++ b/apps/ui/components/Inbox/Inbox.jsx
@@ -1,0 +1,84 @@
+/*
+ * Copyright (C) Balena.io - All Rights Reserved
+ * Unauthorized copying of this file, via any medium is strictly prohibited.
+ * Proprietary and confidential.
+ */
+
+import {
+	circularDeepEqual
+} from 'fast-equals'
+import styled from 'styled-components'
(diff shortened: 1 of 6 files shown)


grahammcculloch (changes requested):
I looked carefully through all the code but can't say I definitely have a handle on all of it. So feel free to set me straight if I've got the wrong end of the stick somewhere. But here's some comments anyway :-)

grahammcculloch (approved):
LGTM - tested locally!

PR #4730, opened 2020-09-08
Add lens-table to support knowledge base view

ljewalsh:
As requested by Hugh: https://jel.ly.fish/a87d37ee-ed8d-4ad6-b16f-f814f1352cd2

I have also run a patch on the card in the production db


StefKors (approved):
LGTM

PR #4781, opened 2020-09-15
ui: Close user menu when navigating

grahammcculloch:
This is a work-around for the fact that the `Link` component stops the click event propagation and overwrites the onClick prop so we can't specifically close the user menu after clicking on one of the links in the user menu!

Change-type: patch

***

test suite!


karaxuna (approved):
LGTM :+1:

PR #4795, opened 2020-09-16
ui: Consolidate View data loading logic

grahammcculloch:
Change-type: patch

***

* As a general pattern, when something changes, call `setState` then call `loadViewWithFilters`
* bind _every_ class method to `this`
* debounce `loadViewWithFilters` method


ljewalsh (approved):
This is a nice cleanup @grahammcculloch

PR #4828, opened 2020-09-22
Improve CRM table

StefKors:
This PR does the following:

## Refactor flow actions and deprecate LinksProvider component

The LinksProvider component was designed to be used with a single SlideInFlowPanel and a single withLinks child component. Using it on a view wil multiple withLinks child components was not possible. This together with the introduction of optional link queries we have no reason to query for card links seperately. We can remove the LinksProvider logic and streamline our SlideInFlowPanel.

This PR also refactors the object stucture to be referenced on the channel target instead of just the card id. With this we can use a single SlideInFlowPanel component with multiple components controlling it's state.

***

test suite!
Screen Shot 2020-09-22 at 6 22 04 PM

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -16,7 +16,8 @@
 	scrub \
 	clean-front \
 	clean-github \
-	npm-install
+	npm-install \
+	npm-ci
 
 # See https://stackoverflow.com/a/18137056
 MAKEFILE_PATH := $(dir $(abspath $(lastword $(MAKEFILE_LIST))))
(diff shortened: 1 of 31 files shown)


grahammcculloch (changes requested):
Not sure of the exact solution for the 'one SlideInFlowPanel but multiple cards' problem. That's the main thing I think we need to sort out on this branch. Otherwise just a bunch of minor React-y things!

PR #4830, opened 2020-09-22
Use snippet for user feedback, add extra views, add chart lens on views

mbalamat:
All of them are 4
Screenshot 2020-09-23 at 12 59 18 AM
I "curated" the 3 of them and it correctly shows 3
Screenshot 2020-09-23 at 1 00 20 AM
And there is only the "un-curated" one left in the "New" User Feedback
Screenshot 2020-09-23 at 1 00 46 AM

### DIFF
--- a/apps/server/default-cards/balena/view-all-users-feedback.json
+++ b/apps/server/default-cards/balena/view-all-users-feedback.json
@@ -24,7 +24,8 @@
     ],
     "lenses": [
       "lens-list",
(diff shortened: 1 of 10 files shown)


grahammcculloch (changes requested):
I may be missing something but it looks like this is the PR where we should remove old/unused views and the e2e tests that go with them.

grahammcculloch (approved):

PR #4931, opened 2020-10-14
ui: Add checkboxes and multi-card actions to table lens

grahammcculloch:
Change-type: minor

***

Multi-Select-Actions

This PR adds checkboxes to the Table Lens. Selecting one or more items in the table enables the 'Actions' DropDownButton. Currently, there are two actions supported:
1. Link to existing element
2. Create a new element to link to

Notes:
* The intention is that these actions will become 'generic card actions' so the list of `ActionLinks` will be automatically generated in the future. But for now we 'hard-code' these two actions.
* Items remain checked after an action is executed on them
* Items remain checked when you change page. TBD: is this desired behaviour? I've tried to allude to it by using the number of checked items in the label of the `DropDownButton` (e.g. 'With 3 selected')

I've added unit tests for the two actions. Note - it's difficult to test the display of the LinkModal itself as it is created in a portal outside of the rendered component.

### DIFF
--- a/apps/server/default-cards/balena/view-all-brainstorm-calls.json
+++ b/apps/server/default-cards/balena/view-all-brainstorm-calls.json
@@ -30,6 +30,7 @@
     ],
     "lenses": [
       "lens-list",
+      "lens-table",
       "lens-kanban"
     ]
 	}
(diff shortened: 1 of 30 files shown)


LucianBuzzo (approved):
Nice!

PR #4972, opened 2020-10-19
Split up package.json - apps/ui

joshbwlng:
Change-type: patch

***

Break bits of package.json out to `apps/ui/package.json`.
This is a prerequisite for livepush and will speed up image builds as only the packages necessary for `apps/ui` will be installed for the `ui` image.


grahammcculloch (changes requested):
Looks ok to me (although I haven't tested it). But it looks like the various make commands that have a `cd` command will leave you in the last sub-directory that the command `cd`s into. If this is the case can you update the makefile to cd back to the root directory after each of these? If it _isn't_ the case, just let me know and I'll approve this PR!

grahammcculloch (approved):

PR #4993, opened 2020-10-22
Fix webpack config for production

joshbwlng:
Change-type: patch

***

Config options for production were lost during webpack config merges yesterday, bringing them back in.


grahammcculloch (approved):
Looks promising!

Issue #3124, opened 2020-02-18
Implement a Contribution Score Card


grahammcculloch:
Contribution score card, for storing data on agent
* This can allow us to easily have different contribution scores per week etc
* Score cards are linked to users
* The score card just needs a field to contain the score value
* Initially the score can be a simple number generated by the schedule
* The score card indicates that the user is enrolled on support, this is important because not every user is a support agent

PR #5013, opened 2020-10-26
WIP: score contribution + render links in json schema renderer

StefKors:
This implements a basic version of the support contribution card described here: https://github.com/product-os/jellyfish/issues/3124

Depends-on: https://github.com/product-os/jellyfish-client-sdk/pull/80
Change-type: minor

***

test suite!


grahammcculloch (changes requested):
Left a few comments.

BTW it doesn't have to be in this PR but we should add suitable restrictions to the roles cards to prevent `regular` users from being able to see (and therefore edit) support-contribution-cards. We're still waiting on a better write permissions system to do this properly. In the meantime though we perhaps need a new role for support auditors? cc @LucianBuzzo

grahammcculloch (changes requested):
Just spotted a couple small things.

PR #5015, opened 2020-10-27
Add eslint to apps/ui

joshbwlng:
Change-type: patch

***

test suite!

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -295,7 +295,7 @@ postgres_data:
 	initdb --locale=en_US.UTF8 --pgdata $@
 
 lint:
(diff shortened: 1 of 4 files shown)


grahammcculloch (changes requested):
So... if we want to be able to call `make lint` from the apps/* directory and have it run eslint we should do this for all apps (i.e. server, action-server and livechat as well). And, as we already have a `lint` command defined in the Makefiles of the various apps I suggest we should migrate _all_ eslinting to these Makefiles (i.e. make the same changes in this PR for action-server, server and livechat and modify the top-level eslint call to only check `scripts` and `test`).

grahammcculloch (dismissed):
Ignore my previous comments - the other apps will be handled in separate PRs!

ghost (approved):

PR #5031, opened 2020-10-29
Fix upsert of messages in inbox

ljewalsh:
Turns out we were importing the update function from immutability-helpers incorrectly. Since we still had the append event, this wasn't obvious to us since we were still getting messages in the inbox. There were, however, errors being thrown all the time in the console.

I did some clean up as part of this PR. Changed the update function to an upsert since this is what it is doing. Also extracted the append function so we can reuse it.


grahammcculloch (approved):
Works for me locally 👍 

I just fixed a small lint error. Reminds me it would be nice to run balena-lint as a pre-commit hook on the Jellyfish repo!

PR #5041, opened 2020-10-30
Livechat: Add redirect logs

karaxuna:
Change-type: minor

***

The user has livechat problem which I found hard to debug (https://jel.ly.fish/997bb386-8158-4488-9670-9a4559ad08b5), adding this logs may help me find the problem.


grahammcculloch (approved):
LGTM - but do check out my couple of comments.

PR #5111, opened 2020-11-06
Update the reason in the reopen and summary triggered actions to align with new designs for updates in the Timeline component

ljewalsh:
This doesn't perfectly align with the designs but we would need to make changes to how we store the reason (at the moment it's stored as a string in the name field of the resulting update card. This gets us 90% of the way there

Screenshot 2020-11-06 at 15 50 29

Change-type: patch
Signed-off-by: Lucy-Jane Walsh <Lucy-Jane@balena.io>

***

test suite!


StefKors (approved):
LGTM

PR #5126, opened 2020-11-09
Refactor: Split out stream events

StefKors:
This splits out the streaming logic it separate files. It's a first pass to clean up the streaming logic for readability

Change-type: minor

### DIFF
--- a/apps/ui/lib/core/store/actioncreators/index.js
+++ b/apps/ui/lib/core/store/actioncreators/index.js
@@ -14,9 +14,6 @@ import merge from 'deepmerge'
 import {
 	once
 } from 'events'
-import {
-	push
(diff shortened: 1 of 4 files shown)


grahammcculloch (changes requested):
Seems like a good change - will definitely make it easier to unit-test this functionality in the future. 

I think when you make a refactor like this it is worth adding some doc comments, e.g. above the new `streamTyping` and `streamUpdate` methods. Now that they are in their own file, the context is lost and it's not so clear just by looking at those methods what their purpose is and when they might be used and even what the arguments should be (e.g. what is `types` or `payload`).

grahammcculloch (approved):

PR #5151, opened 2020-11-12
Add a new mixin that will add pipeline item attributes

LucianBuzzo:
At the moment this is just the status enum and slicing.

Change-type: patch

***

test suite!


grahammcculloch (approved):
Nice :-)

PR #5167, opened 2020-11-14
Add Livepush development flow

joshbwlng:
Change-type: patch

***

ui-components-livepush

### App source
App source (`apps/<app-name>/lib/`) syncing is done by Livepush:
- Copied to `/usr/src/jellyfish/apps/<app-name>` in each app `Dockerfile`s
- Kept up to date by Livepush
- `nodemon` and `webpack-dev-server` are configured to watch these files and hot reload on updates

### Library source
Library source (`<library-name>/lib/`) syncing is done by Livepush:
- The library branch you want to work on needs to be cloned and checked out under `.libs/<library-name>`
- Copied to `/usr/src/jellyfish/apps/<app-name>/node_modules/@balena` in app `Dockefile`s
- Default packages are installed using `npm ci` at an earlier layer, this copy overwrites those packages
- Kept up to date by Livepush
- `nodemon` and `webpack-dev-server` are configured to watch these files and hot reload on updates

### App dependencies
App-level additions and removals are handled by Livepush alone as well. Install/remove a dependency like you normally would locally and this will trigger that app's image to be rebuilt from the `npm ci` layer.

### Library dependencies
Library-level additions and removals are a bit different and less "automatic". For example, to add/remove a dependency for `jellyfish-worker`:
- `cd .libs/jellyfish-worker`
- `npm install/uninstall <dependency-name>`
- `cd ../..`
- `make deploy-worker`

This creates a tarball of the current source of `.libs/jellyfish-worker` into what can effectively be thought of as a local beta package. This tarball is then copied into whichever apps use `@balena/jellyfish-worker` (by checking each app's `package-lock.json`). This copy will trigger Livepush to rebuild the image from the layer just before the dev comment is executed. This process should take less than 30 seconds for the `ui` service and usually around 10 seconds or so for backend services.

## Running
1. Set up balena device and set to local mode
2. Make sure `jel.ly.fish.local` resolves to device local IP (might need to get IP with `balena scan` and map in hosts)
3. add `- NPM_TOKEN=xxx` with your token to `.balena/belena.yml` under `build-variables.global`
4. `cd` to `.libs`, clone library you are going to work on, and checkout your branch
5. Go back to repo root and execute `make push`

## Troubleshooting
- Tail individual service logs with `balena logs jel.ly.fish.local --service <name>`
- Log into device with `make ssh`, which allows you to then:
  - Check running containers with `balena ps`
  - Enter running container with `balena exec -ti <name> bash`
  - Check container logs directly from within device using `balena logs <name>`

## ToDo
- [ ] Add docs
- Find better way to pass `NPM_TOKEN` as a build argument
  - This is currently a limitation of balenaCLI, as it only allows for passing build args as command line arguments for `balena build` and not `balena push` (https://github.com/balena-io/balena-cli/issues/1053)
- Add sidecar to enable balena device-side testing
  - This will be done in a subsequent PR

## Tests
- [x] Local development still works as expected
- [x] Tests pass in CI
- [x] Updating app source reloads service without image rebuild
- [x] Updating app dependencies causes rebuild from `npm ci` layer
- [x] Updating library source reloads service without image rebuild
- [x] Deploying library tarball rebuilds image from `/packages/` `COPY` layer


LucianBuzzo (approved):
🎆 🥳 🎈

PR #5177, opened 2020-11-16
Optimize webpack dev configs

joshbwlng:
Change-type: patch

***

Tweak webpack `optimization` config used during `make dev-ui` and `make dev-livechat`. Pulled tips from:
- https://webpack.js.org/guides/caching/
- https://blog.box.com/how-we-improved-webpack-build-performance-95

Also think it would be better to stop using `BundleAnalyzerPlugin`? If it's something that the frontend developers make use of quite often, it should stay, otherwise I doubt we need to generate the report every time.

**The point of this PR is to find a way to optimize webpack builds during development so builds don't die from OOM errors so if anyone has other ideas on how to do this don't hesitate to let me know or push commits!**

### DIFF
--- a/apps/livechat/lib/webpack.config.js
+++ b/apps/livechat/lib/webpack.config.js
@@ -40,7 +40,8 @@ const config = {
 					{
 						loader: 'babel-loader',
 						options: {
-							presets: [ '@babel/preset-react' ]
+							presets: [ '@babel/preset-react' ],
+							cacheDirectory: true
 						}
 					}
 				]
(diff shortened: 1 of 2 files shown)


grahammcculloch (changes requested):
I've experimented with numerous variations of optimization settings in the branch and I can't get it to produce anything as sensible as what was on master! With the settings in this PR, we end up with a 43MB vendors.js file (as opposed to master where the largest file is main.js at 23MB - still way too big IMHO).

I think if you can verify that `cacheDirectory` is working (see my inline comment about that) I would recommend just pushing that change and reverting any changes to do with the `optimization` settings for now.

grahammcculloch (dismissed):
Let's do this!

grahammcculloch (approved):

Issue #4625, opened 2020-08-20
Unread messages indicator shows unread messages even though inbox is empty


grahammcculloch:
## Expected Behavior
If the inbox is empty (no unread messages), the unread indicator should not display any messages.

## Actual Behavior
Sometimes (it's not clear what the steps to reproduce this are) the unread messages indicator remains displaying that there are one or more unread messages. But if you go to the inbox, no unread messages are shown. Refreshing the page clears the unread messages indicator.

PR #5182, opened 2020-11-16
Test: mentions count should be equal to messages in inbox

StefKors:
This test checks if the inbox and homechannel mentions have the same count. It's part of investigating issue #4625 

Related-to: https://github.com/product-os/jellyfish/issues/4625
Change-type: patch

***

test suite!


grahammcculloch (changes requested):
I think chat.spec.js would actually be a better place for this kind of test. It already creates two users and two windows so you don't need to worry about logging in and out etc. To be honest you can also simplify it a lot by just using the SDK to create messages rather than exercising the UI. This test is not, after all, testing the message creation UI but only the mentions count and inbox behaviour. So we should restrict any other code to its simplest form - in this case using the SDK.

grahammcculloch (approved):

PR #5208, opened 2020-11-19
Allow setting FILES for make test in apps/ui

joshbwlng:
Change-type: patch

***

Allow setting `FILES` when running `make test` under `apps/ui`, defaulting to `"'./{lib,test}/**/*.spec.{js,jsx}'"`.
Also setting a default value for `FILES` in `apps/server/Makefile` along with a couple convenience make commands:
- `make test-unit`
- `make test-integration`

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -318,7 +318,7 @@ test: scrub
 
 test-unit:
 	cd apps/ui && make test
-	cd apps/server && FILES="'./test/unit/**/*.spec.js'" make test
+	cd apps/server && make test-unit
 
 test-integration:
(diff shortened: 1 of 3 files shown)


grahammcculloch (changes requested):
This looks fine but would be really helpful to add the same feature to apps/server as well. And that should be a bit more involved as there are both integration and unit tests in apps/server. But currently, I can't do something like the following:
```
cd apps/server
make test FILES=test/unit/http/facades/view.spec.js
```

grahammcculloch (approved):

PR #5256, opened 2020-11-27
Revert "Add Livepush development flow"

LucianBuzzo:
Reverts product-os/jellyfish#5167


StefKors (approved):

StefKors (changes requested):
removing review so we don't auto merge

PR #5260, opened 2020-11-27
Rating

karaxuna:
Change-type: minor

***

tmp

### DIFF
--- a/apps/server/lib/default-cards/balena/view-all-forum-threads.json
+++ b/apps/server/lib/default-cards/balena/view-all-forum-threads.json
@@ -34,7 +34,9 @@
                     "message@1.0.0",
                     "create@1.0.0",
                     "whisper@1.0.0",
-                    "update@1.0.0"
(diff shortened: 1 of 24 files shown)


grahammcculloch (commented):
This is looking good 👍 - nice work. I've left a couple comments about things I've spotted.

Feels like we should invest some time abstracting out helpers for 'timeline' cards. There's just a lot of places in the codebase where we are listing out card types. It started with message and whisper and now its message, whisper, summary and rating. It would be very easy to miss a place where we need to add a card type to a list and this seems dangerous.

grahammcculloch (commented):
@karaxuna I'm curious about the two changes in `macros.js`. These will affect other tests and, whilst the tests seem to all be passing, I wonder if these changes are really required. Perhaps you've found they are an improvement? If so, could you comment on how/why these changes were made.

grahammcculloch (approved):
Ok - seems good to me!

PR #5281, opened 2020-12-04
Allow UI containers to be parameterized at runtime

LucianBuzzo:
See [this blog post](https://www.freecodecamp.org/news/how-to-implement-runtime-environment-variables-with-create-react-app-docker-and-nginx-7f9d42a91d70/) for a good explanation of the problem we want to solve and how the solution works.

Relies on https://github.com/product-os/jellyfish/pull/5366 so that the required env vars are in place on our production deployments.

test suite!


grahammcculloch (commented):
Left one question about default values

ghost (approved):

PR #5330, opened 2020-12-13
Add loop and transformer worker role.

LucianBuzzo:
Loops have the same permissions as superuser. Transformer worker permissions are modelled off the user-community ones,
with the exception of being able to create session cards for any user and not having permissions on views

Change-type: patch
Signed-off-by: Giovanni Garufi <giovanni@balena.io>

***

test suite!


grahammcculloch (commented):
Had a question about reading active=false cards. But probably not a show-blocker at this stage if we want to get this PR merged.

ghost (approved):

PR #5405, opened 2020-12-23
Remove home channel

karaxuna:
Change-type: minor

***

See conversation: http://localhost:8080/320f5771-ef6f-4b06-8831-add26e209e11

Note that with this modification ui is bootstrapped with `user-guest` is user is not logged in. Ui is rendered starting from `misc/User` lens.


joshbwlng:
@karaxuna Do we still need this?

PR #5444, opened 2021-01-06
Replace balena-redis with official redis images

joshbwlng:
Change-type: patch

***

Replace `balena-redis` image with official `redis` image, as we are deprecating [`balena-redis`](https://github.com/product-os/balena-redis). The biggest difference with these changes is that the `balena-redis` image we were using in compose and production was based on Redis 5.0.3 while this PR will change this to the most recent stable image based on 6.0.9.


cmfcruz (dismissed):
Looks safe to upgrade from redis 5.0.3 to 6.0.9. Lez go!

cmfcruz (approved):
Safe to upgrade from Redis 5.0.3 to 6.0.9. Let's have this out.

PR #5506, opened 2021-01-15
Expand links directly in stream

StefKors:
As far as possible, all data fetched from the server should be fetched by a stream.

- This will hugely improve the UX as front-end data will automatically be kept in sync with backend data.
- It will simplify the front-end code significantly as there will be no requirements for manual updating/syncing across sibling components.

The main risk is that the backend will be overloaded with the increased size of stream queries.


grahammcculloch (changes requested):
Stef - I think you still need to handle [the query that is generated by ` Timeline` when paginating](https://github.com/product-os/jellyfish-ui-components/blob/d0cd4cf7c93a99420fc6f2b2c7694e84f4e0f498/lib/Timeline/index.jsx#L348-L372).

Also left one comment about filtering by `data.from` when fetching all links.

PR #5533, opened 2021-01-20
Use snippet view mode in SingleCard snippet lens

grahammcculloch:
Add optional `viewMode` prop to the `CardFields` component.

Change-type: minor
***

This will reduce the amount of data displayed for each card in various views. But this should be an improvement overall. As and when we think more detail is required in the list view, we can just update the relevant card type's `snippet` view mode to include the fields that are required.


StefKors (commented):
@grahammcculloch Do you have an idea which cards will lose displayed fields? I'm wondering if this PR already should update those cards to display their information correctly?

StefKors (approved):
Awesome thanks @grahammcculloch

PR #5580, opened 2021-01-26
Stop using SupportThreadsToAudit lens for viewing non-support-thread cards

grahammcculloch:
This PR fixes a super-old bug where `lens-support-threads-to-audit` was incorrectly being used to render lists of cards that were not support-threads or sales-threads.

Before:
Brainstorm Calls Topic Segment - before

After:
Brainstorm Calls Topic Segment - after

The easy part of fixing this bug was adding details to the `filter` property for `lens-support-threads` and `lens-support-threads-to-audit`. 

Significantly, the `Segment` component now typically uses the `lens-list` to render the list of cards (instead of `lens-support-threads-to-audit`!). Before this was possible I had to make a few tweaks to `Segment` and `List` to get them to play nicely together. The exception is, of course, tabs displaying support-threads or sales-threads. These are now rendered using the `lens-support-threads` (instead of `lens-support-threads-to-audit` - which requires that all threads are closed).

The code changes are separated out into several distinct commits so you can see _why_ each change was made. 

### Note
One slightly hacky part of all this is that the `List` component auto-sizes based on it's parents size. This is fine when the list is located within a `flex: 1` div that will always have a relatively large height (at least several times the height of an item in the list). But if the `List` component is in a 'normal' layout (i.e. not expanding to fill a flexbox container), we need to specify a hard-coded min height to ensure it renders with that height and doesn't collapse to zero height. What would be much better is if the `List` could increase in size based on it's item and you just specify an optional _max_ height for the `List` instead. PRs to fix this are welcome! 😆

### DIFF
--- a/apps/ui/lib/lens/common/Segment.jsx
+++ b/apps/ui/lib/lens/common/Segment.jsx
@@ -183,11 +183,16 @@ class Segment extends React.Component {
 
 		return (
 			<Flex flexDirection='column' style={{
-				height: '100%'
+				flex: 1
 			}}>
-				<Box flex={1} style={{
+				<Box flex={1} mt={3} style={{
 					minHeight: 0
 				}}>
+					{results.length === 0 && (
+						<Box px={3}>
+							<strong>No results found</strong>
+						</Box>
+					)}
 					{Boolean(results.length) && (
 						<lens.data.renderer
 							tail={results}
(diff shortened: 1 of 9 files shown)


StefKors (approved):
LGTM

@grahammcculloch About the note: I think this lens tab is quite messy. Not just in stying but also the layout could use improvements. Holding off until auto UI is probably fine for now. Otherwise I'd suggest we move some elements around so we can render a simple list with less cascading flex styles.

PR #5591, opened 2021-01-27
Fix link on SingleCard snippet lens

grahammcculloch:
The link should append the target channel to the current channel (i.e. replacing any existing content to the right of the current channel).

This change introduces a React context which supplies the channel value. This is useful for any components that need to know what channel they are being displayed in.

Change-type: minor

***

Fixes #5582

## Before

With a brainstorm call open, I choose a topic which opens on the right hand side of the screen (perfect), but when I choose a second topic, the first topic moves to the left and I end up with two topics open simultaneously. 

## After

With a brainstorm call open, I select a topic and have it appear on the right hand side of the screen, then choose a different topic and have the right hand side view switch to the new topic.


karaxuna (approved):
LGTM

PR #5606, opened 2021-01-29
Share Puppeteer browser options

joshbwlng:
Change-type: patch

***

Simple PR that shares the options used to create Puppeteer instances during tests in order to reduce duplicate code. The options are the exact same for both `e2e-ui` and `e2e-livechat` tests.


grahammcculloch (approved):
LGTM 

(Just noting that we'll have to refactor e2e helper code if we ever split the livechat and UI applications into their own repos. But that problem existed before this PR!)

PR #5696, opened 2021-02-11
Automatically select list lens options

grahammcculloch:
Rather than getting the list of lenses from the hard-coded values in the type card, lenses will now be fetched based on their validity and scoring. All list lenses that are valid for the given data are grouped by lens 'icon' and the top-scoring valid lens for each icon is returned.

Change-type: minor

***


StefKors (commented):
@grahammcculloch Looks good, small nit:

- I don't understand how the lens scoring works, could you explain it a bit more?
- Should we have a test that demonstrates lens scoring?
- Do we have big changes in lens selection? or does that stay the same for most views?

LucianBuzzo (approved):

StefKors (approved):
Thanks @LucianBuzzo :shipit:

PR #5771, opened 2021-02-18
Refactor action facade

karaxuna:
Change-type: minor

***

`ActionFacade.processAction` async function does not throw an error when it happens, instead it returns object `{ error: boolean, data: any }`. This refactor makes `processAction` throw if `result.error` is true with `errio.fromObject(result.data)` error or return `result.data` data if there's no error.

### DIFF
--- a/apps/server/lib/http/facades/action.js
+++ b/apps/server/lib/http/facades/action.js
@@ -5,6 +5,7 @@
  */
 
 const _ = require('lodash')
+const errio = require('errio')
 const logger = require('@balena/jellyfish-logger').getLogger(__filename)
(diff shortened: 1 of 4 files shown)


grahammcculloch (commented):
@karaxuna to be honest I don't feel confident approving this as it seems to make a _lot_ of changes to error codes and error-handling, with very few changes/additions to tests. Which I suspect is because we don't have enough existing tests for error conditions. 

What might help is if you annotate this PR with some inline comments explaining what each set of changes is for or the significance of them.

grahammcculloch (approved):
Ok - I think this is fine!

PR #5813, opened 2021-02-24
clarify use of the chat widget

Hades32:
I was surprised what that button actually does. The UX is still problematic (+ for private?), but this should improve the first-time user experience a bit.


grahammcculloch (approved):
Thanks @Hades32  - nice simple improvement :-)

PR #5898, opened 2021-03-04
Add oauth env vars for ci

karaxuna:
Change-type: minor

***

test suite!


ghost (approved):

Hades32 (commented):
tests seem to be expecting a different client-id

joshbwlng (changes requested):
Need to remove hard coded values and instead add var names only. Values should be set in the AWS parameter store as so to not break our dev .env setup.

joshbwlng (approved):

PR #5951, opened 2021-03-11
User profile: Allow copying own session token

karaxuna:
Change-type: minor

***

Screen Shot 2021-03-11 at 7 19 02 PM


LucianBuzzo (approved):
Really nice 👍 Small nit comment that can be adressed in a follow up PR

PR #5955, opened 2021-03-12
Keep MyUser as presentational component

grahammcculloch:
In general we should try and keep HOCs out of presentational component files. This makes testing the presentational component much easier.

Change-type: patch

***


karaxuna (approved):
Nice ^_^

PR #5991, opened 2021-03-16
JellyfishUser widget should cope with null user

grahammcculloch:
Change-type: patch

***

While the user card is being fetched, we need to display a sensible link while `user` is null/undefined.


karaxuna (approved):
LGTM

PR #6130, opened 2021-04-08
[TEST] Add cache_from instruction in compose file

nazrhom:
Change-type: patch

***

test suite!


LucianBuzzo (dismissed):
See my comment. To support this we need to improve balena push support as well

PR #6355, opened 2021-05-09
Fix OmniSearch lens search types logic

joshbwlng:
Change-type: patch

***

The current implementation of the logic that finds contract types that contain full-text searchable fields doesn't account for slightly complex field definitions such as `oneOf`. I don't believe this currently affects any contracts, but could in the future if we add a contract that has a single full-text searchable field that is defined with `oneOf`, etc. These changes modify that logic to search deeper within field children instead of assuming `fullTextSearch: true` will exist at the top level.

Found this while looking for where the omnisearch query is generated on the front-end as it doesn't include fields such as the `user` contract's `data.email`. I'm *assuming* this is because `data.email` is defined with `oneOf`.

## Issues
- Would like to do this without adding a new dependency (currently adding `deepdash`)

## Example
```
const _ = require('lodash')
const findPathDeep = require('deepdash/findPathDeep')

const schema = {
	email: {
		oneOf: [
			{
				type: "array",
				fullTextSearch: true
			},
			{
				type: "string",
				fullTextSearch: true
			}
		]
	}
}

console.log('_.find:', Boolean(_.find(schema, {
	fullTextSearch: true
})))

console.log('findPathDeep:', Boolean(findPathDeep(schema, (v, k) => {
	return Boolean(k === 'fullTextSearch' && v)
})))
```

Output:
```
$ node find.js
_.find: false
findPathDeep: true
```

### DIFF
--- a/apps/ui/lib/lens/misc/OmniSearch.tsx
+++ b/apps/ui/lib/lens/misc/OmniSearch.tsx
@@ -4,6 +4,7 @@
  * Proprietary and confidential.
  */
 
+import findPathDeep from 'deepdash/findPathDeep';
 import React from 'react';
 import _ from 'lodash';
 import { v4 as uuid } from 'uuid';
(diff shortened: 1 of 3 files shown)


grahammcculloch (commented):
This is a good catch and something that's definitely worth fixing. 

I'm just not sure the solution here will actually fix it - see comment.

LucianBuzzo (approved):

PR #6558, opened 2021-06-02
Add k8s checks to action-server

joshbwlng:
Change-type: minor

***

Depends on: https://github.com/product-os/jellyfish-environment/pull/225

This PR adds an HTTP server with `/startup` paths for Kubernetes checks for `action-server` and `tick` servers. The ports used by these HTTP servers are `HTTP_TICK_PORT` and `HTTP_WORKER_PORT` as defined in the `jellyfish-environment` PR linked to above. I have followed basically the same pattern as the setup used in `apps/server` to make it easier when we finally get around to merging the two backend services. It should also be noted that while the `tick` server is not as important as the `action-server` in regards to checking readiness, the tick server uses the same bootstrap function and should therefore also be subject to the same readiness/liveness checks.

Have tested locally and everything seems to be fine, but am unable to confirm that my Kubernetes config updates are valid for production, so would appreciate some extra attention there.


cmfcruz (dismissed):

Hades32 (changes requested):

Hades32 (dismissed):
LGTM! Only added small optional nit about named ports

grahammcculloch (commented):
Are the secrets in place on production for these two new environment variables?

grahammcculloch (commented):
@joshbwlng if you can just go through existing conversations and mark them as resolved I will be confident to approve this PR. I'll probably dismiss Carlo's review as it looks like you've already resolved his comments.

grahammcculloch (approved):
Let's do this!

PR #6583, opened 2021-06-08
Display version of cards in UI

Hades32:
show slug+version in the small type header and display the version in lists IF it's not a 1.0.0 version

Edit: note, the versions you see in the screenshot below are now only displayed for non 1.0.0 versions
image

Resolves https://github.com/product-os/jellyfish/issues/6399

### DIFF
--- a/apps/ui/lib/layouts/CardLayout/index.tsx
+++ b/apps/ui/lib/layouts/CardLayout/index.tsx
@@ -36,19 +36,21 @@ const CardLayout = (props) => {
 
 	const typeBase = card.type && card.type.split('@')[0];
 
-	const typeName = _.get(
+	const typeContract =
 		_.find(types, {
 			slug: typeBase,
-		}),
-		['name'],
-		null,
-	);
+		}) || {};
+	const typeName =
+		!typeContract.version || typeContract.version === '1.0.0'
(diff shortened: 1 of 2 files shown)


grahammcculloch (dismissed):
@Hades32 it looks like this change is mainly beneficial for transformers. I'd be wary of modifying the UI for _every_ contract type in this way. Certainly at the moment the user is not concerned about versioning for any other contract type. Can I suggest making use of the `snippet` UI schema  on the `transformer` contract to expose the version field? That way we won't unnecessarily display the version on _every_ contract type.

As far as the `CardLayout` goes... can we compromise and display the slug and version in a tooltip or something? Again, adding the slug and version to the header of _every_ contract clutters the UI at very little benefit to the user (unless it is a transformer contract). Alternatively you could also expose the `version` field in the `fields` UI schema for a `transformer` contract. 

In general, it's probably worth adding me as a reviewer for any UI PRs as I'm closest to that code.

LucianBuzzo (approved):
Looks good to me 👍 
We might want to tweak this in the future as part of a larger design review, but its easily changed. I think the small decision to only show it if the version *is not* 1.0.0 neatly solves a lot of display concerns: good work! 😄

PR #6614, opened 2021-06-10
Fix bugged date comparison in Front inbox cleanup script

LucianBuzzo:
When check front conversations, we were previously comparing the wrong unit
level: seconds to milliseconds.

This change also switches to the slightly more correct /readiness
endpoint when waiting for the API to start up in e2e tests.


grahammcculloch (approved):
Nice catch! 

I actually favour including the time unit in the variable name to avoid bugs just like this. E.g. `created_at_s`.

PR #6622, opened 2021-06-11
Expose all relationships in tabs

grahammcculloch:
Abstract this functionality into a reusable RelationshipsTab component.

Also refactored the CreateLens and EditLens components to always display the submit and cancel buttons at the bottom and make the form itself scrollable. This should improve the UX as the user can always quickly press cancel or submit and it's clear (from the submit disabled state) whether the form data is valid.

Change-type: minor

***

Note: 'Custom' meta relationships defined in contract cards (e.g. ['Sales' in the user contract](https://github.com/product-os/jellyfish-core/blob/2b9982138fd5085fc108958df4cc57deae538102/lib/cards/user.ts#L350-L385)) are still rendered as their own tabs.

A follow-up task will be to remove the `withRelationships` mixin and all 'link' relationships that are explicitly defined in contract cards. The `Owned Conversations` custom query relationship can also be removed as you can access owned support threads and sales threads through the new 'Links' tab now.
 
Relationships Tab

Basic unit test added and a more comprehensive e2e test (as the `Select` component proved hard to unit test!)

### DIFF
--- a/apps/ui/lib/index.tsx
+++ b/apps/ui/lib/index.tsx
@@ -80,6 +80,10 @@ const customTheme: any = {
 		background: '#fff',
 		border: '#eee',
 	},
+	tab: {
+		// Keep tab height consistent with height of Select component
+		extend: `${Theme.tab.extend}; height: 32px;`,
+	},
 };
 
 const widgets: any = {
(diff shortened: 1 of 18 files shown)


LucianBuzzo (approved):
LGTM - any idea what we can do with these "custom" tabs to make them more generic. I really want to support traversing from contacts to sales/support threads easily

PR #6665, opened 2021-06-16
Fix pgp import

karaxuna:
Change-type: patch

***

After rewriting jellyfish-core to ts, module path is `/build` instead of `/lib`.

test suite!


grahammcculloch (approved):
Ideally we should export pgp from the top level index in jellyfish-core if it is used outside of the package.

PR #6671, opened 2021-06-17
Livepush: Ignore lib TypeScript source

joshbwlng:
Change-type: patch

***

These changes should stop TypeScript source located under libs from being watched by Livepush. `build/` is still watched and copied into containers. The only exception to this is `jellyfish-plugin-default` as it has not yet been converted to TypeScript. The last line referring to `jellyfish-plugin-default` should be removed once it has been converted to TypeScript.

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -12,3 +12,6 @@ postgres_data
 **/package*.json.*
 .libs/*/package*.json
 .libs/*/*.tgz*
+.libs/*/lib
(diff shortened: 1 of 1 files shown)


Hades32 (approved):
LGTM. How do you handle building stuff in the lib folder? I'm currently starting onw `tsc -w` in each of my `.lib` modules which feels ineffecient...

PR #6741, opened 2021-06-23
Exclude index.html and build folder output from service worker precache manifest

grahammcculloch:
The build folder output (as opposed to the dist folder) contains only TypeScript definition files.

Excluding index.html will hopefully fix the problem where some people are seeing blank screens on start-up.

Change-type: minor
***

Fixes #3887

Note - I think since we migrated the UI app to TypeScript the service worker caching has been more broken than we realised. It was trying to cache all the `apps/ui/build/*.d.ts` TypeScript definition files generated as part of the build process. This resulted in run-time workbox errors (which don't show up in production as we turned off workbox logging). So I've also excluded these files from the precache manifest. 

I've tested this locally and it correctly caches expected files (e.g. main.js, css files, etc) but not the index.html file. When I make a code change, rebuild and then refresh the browser, the updated index.html is loaded, the updated main.js file is fetched from the network and cached. The 'New version of Jellyfish available. Update now?` popup shows. When I click 'OK', the browser tab refreshes and this time the main.js file is fetched from cache.

### DIFF
--- a/apps/ui/.gitignore
+++ b/apps/ui/.gitignore
@@ -4,3 +4,6 @@ env-config.js
 # the command 'npm run doc' and committed to the root of the 
 # 'docs' branch.
 docs/
+
+# Ignore local build output (typescript definition files)
+build/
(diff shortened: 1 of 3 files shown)


Hades32 (approved):
Cool! Hope it helps! 

Although, TBH I don't understand why we're using service worker request caching at all. Can we really cache much? And a offline JF version probably not coming soon 😬

PR #6861, opened 2021-07-07
Remove frontend Renovate group

joshbwlng:
Change-type: patch

***

@grahammcculloch I believe we can remove this group now that `jellyfish-ui-components` and `jellyfish-chat-widget` no longer have `package-lock.json` files? (aka Shouldn't run into the multiple versions of react problem any longer.)


grahammcculloch (dismissed):
Yay!

grahammcculloch (approved):

PR #6866, opened 2021-07-07
Improve livechat tests

karaxuna:
Change-type: minor

***

Test livechat exactly how it is used in balena dashboard. Add test for notifications event emitted on parent window.

test suite!


grahammcculloch (approved):
Ok - LGTM

PR #6885, opened 2021-07-08
Merge action server contract loading into a single stream and query

LucianBuzzo:
This change also updates type contracts in realtime, which previously
did not happen.

Change-type: patch

***

test suite!


karaxuna (approved):
Nice, just needs prettifying.

PR #6898, opened 2021-07-09
Accept navigation event from parent window

karaxuna:
Change-type: major

***

Livechat should be able to navigate to url passed by parent window emitting `navigate` event with `livechatIframe.postMessage`. With previous structure I could not change livechat component property when I received navigate event, so I had to move window's message event listener into the component, but this introduced new challenge, because livechat react component may not be initialized when it received `init` event, so I had to remove this event and initialize livechat on page load (I think it should have worked this way from beginning). Dashboard already has this scenario implemented, see: https://github.com/balena-io/balena-ui/pull/4628

Flowdock thread that led to this change: https://www.flowdock.com/app/rulemotion/resin-frontend/threads/e83ksByzqoTAfnDFSuhDeftwmVX

test suite!

### DIFF
--- a/apps/livechat/lib/components/AuthenticationTask.jsx
+++ b/apps/livechat/lib/components/AuthenticationTask.jsx
@@ -34,7 +34,7 @@ const authenticate = async ({
 }
 
 export const AuthenticationTask = ({
-	userSlug, oauthUrl, children
+	userSlug, oauthUrl, children, ...rest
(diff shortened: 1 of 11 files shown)


LucianBuzzo (dismissed):
LGTM, also, great write up of the PR! More like this please 😄

ghost (approved):

PR #7041, opened 2021-08-03
Inherit loop when creating new contracts

grahammcculloch:
Inherit loop when creating new contracts

Turns out there's quite a few places in the code where we create new contracts! (We should refactor/abstract this in a future PR).

Change-type: patch
***

Fix linking to new accounts in CRM table

This was sooooo broken before!

Change-type: patch
***


karaxuna (approved):
LGTM

PR #7089, opened 2021-08-17
Add a simple UI for check-runs

LucianBuzzo:
This commit adds a basic UI for check runs that shows all of the
contracts created by transformers from the commit the check run is
attached to. The graph is generated using mermaidJS, so doesn't add any
additional dependencies to the system. In the future I'd like to develop
this concept into a generic graph visualisation features, however there
are behavioural issues that need to be resolved, such as which links you
expand/display and to what depth.

image

Resolves https://github.com/product-os/product-os/issues/1180

### DIFF
--- a/apps/ui/jest.config.js
+++ b/apps/ui/jest.config.js
@@ -14,6 +14,7 @@ module.exports = {
 		env: {}
 	},
 	transform: {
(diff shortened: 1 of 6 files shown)


Hades32 (commented):
apart from my questions above, LGTM

ghost (approved):

PR #7096, opened 2021-08-18
Remove tick server

joshbwlng:
Change-type: major

***

Remove the `tick` server entirely from our backend as (a) we currently do not use it and (b) the intended functionality of `tick` will be handled by the introduction of [scheduled actions](https://jel.ly.fish/88e737da-8692-49c8-bf5a-43d27c6b7a39).


LucianBuzzo (commented):
Yep this looks great! Would you be able to add a paragraph to the commit message explaining why we're removing it and provide a link to the scheduled action improvement?

LucianBuzzo (approved):

PR #7201, opened 2021-09-15
Support authorization with http-only cookie

karaxuna:
Change-type: minor

***

Add new endpoints that do login, logout and refreshToken actions and modify token cookie. This change is compatible with authorizing with Authorization header. Next step is to modify client sdk to use these endpoints, instead of actions directly.

test suite!


joshbwlng:
@karaxuna Do we still need this?

PR #7274, opened 2021-10-07
Switch to esbuild loader for faster UI bundling

LucianBuzzo:
On my local machine this change reduced UI build times from 3m7s to 56s which is a massive improvement.

Change-type: patch

***

test suite!


karaxuna (approved):
LGTM :+1:

PR #7324, opened 2021-10-21
Include md5sum hash in env-config.js filename

joshbwlng:
Change-type: patch

***

We had [an incident](https://www.flowdock.com/app/rulemotion/resin-devops/threads/qhU0CLuemFfHxdrR1dtY4JzUT1M) caused by a broken version of Livechat's `env-config.js` being cached by CloudFlare. We have been able to resolve the problem by manually purging the broken cached version from CloudFlare. This isn't a long term solution however, and in order to force downloads of the most recent version this PR adds in the md5sum hash of the file in the filename. This means `env-config.js` file will now be something like `env-config.f8a5badce1d2a93432bc3a04bdbbcbfb.js` (reference in `index.html` is also updated). This same problem also exists for `apps/ui` so including the same fix for that as well.

Also stop copying `env.sh` to the public html directory. Instead copy to `/tmp` and execute it from there to generate `env-config.<hash>.js`.

### DIFF
--- a/apps/livechat/Dockerfile
+++ b/apps/livechat/Dockerfile
@@ -32,11 +32,11 @@ RUN make build-livechat NODE_ENV=production
 FROM nginx:1.21.3
 WORKDIR /usr/share/nginx/html
 COPY --from=base /usr/src/jellyfish/apps/livechat/dist/livechat /usr/share/nginx/html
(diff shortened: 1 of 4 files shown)


LucianBuzzo (approved):
Good spot @joshbwlng

PR #7348, opened 2021-10-29
Ensure streams are closed correctly when navigating through the UI

LucianBuzzo:
Previously streams would only be closed if the `removeChannel` action
was called explicitly. Now streams are always tidied up after a navigation
event, when the `setChannels` action is called.

Change-type: patch

***

test suite!


karaxuna (approved):
Looks good :+1:

PR #7368, opened 2021-11-02
Katapult: Standardize the labels used for the K8s resources.

cmfcruz:
This is to make sure that the labels used for the resources
follow a standard format.  This also adds necessary labels
for the namespace resource to make sure that ALB ingress
controller detects the new pods and injects the needed
readiness gates.

Change-type: patch
Signed-off-by: Carlo Miguel F. Cruz <carloc@balena.io>

***

test suite!


joshbwlng:
Moved to balena, no longer deploying to k8s

PR #7410, opened 2021-11-12
Remove unneeded env var declarations

joshbwlng:
Change-type: patch

***

We now have most necessary environment variables either defined as secrets or with sane defaults to fall back to in `jellyfish-environment` for development and testing. These changes seem small, but can easily break people's development flows so even if things pass in CI, we need to check if it breaks anything commonly used for local development.


joshbwlng:
Am going to do this in separate smaller PRs to make the transition a bit easier

PR #7413, opened 2021-11-15
Develop with livepush using target field

karaxuna:
Change-type: minor

***

This pr is about using multistage build to build libraries and then copy output to apps. Main dockerfile is generated with script `npm run generate-dockerfile`, which reads modules inside apps and libraries, analyzes their dependencies and generates stages based on this information.

test suite!


robertgzr:
@karaxuna is this what you see?
```
[Info]    Starting build on device 127.0.0.1
[Info]    Docker compose dev overlay detected (docker-compose.dev.yml) - merging.
You must specify either an image or a build
```

PR #7426, opened 2021-11-18
Merge Makefiles

joshbwlng:
Reduce complexity and places in which environment variables
could be set or overridden.

Change-type: patch

***

test suite!


joshbwlng:
@balena-ci rebase

Makefile
```diff
@@ -240,7 +242,7 @@ else
 SCRUB_COMMAND =
 endif
 
-SENTRY_DSN_UI ?=
+SENTRY_DSN_UI ?= https://ff836b1e4abc4d0699bcaaf07ce4ea08@sentry.io/1366139
```
Hades32, line 245: will this send my local issues to the prod sentry instance? Isn't this token secret?

Hades32 (commented):
can't we get rid of all makefiles? I really don't understand why we would need them. Worst case we could put everything in npm run scripts...

Makefile
joshbwlng: @Hades32 We already had this in `apps/livechat/Makefile`, the idea was to move it here for simplicities sake. I wasn't sure if it was supposed to be a secret, but if so we probably need to regenerate it, it's been in clear text in this repo for ages 😓 

I get your point about it affecting local devenvs, we should explicitly set this to nothing for devenvs.

joshbwlng:
| can't we get rid of all makefiles?

@Hades32 That's my plan, eventually get rid of all Makefiles (already did this for all JF lib repos if you haven't noticed). This one is a bit more difficult to drop, but I'm looking to slowly phase it out, especially after Livepush-based development is a bit less broken and people don't need to fallback to local development from time to time.

I guess we could just drop the Makefiles now and use a combination of npm scripts and possibly shell scripts if we need some extra logic for any reason. Might tackle that in an experimental branch next week or the next hack Friday. For now though, you can think of this PR as the first step in removing all Makefiles.

Makefile
joshbwlng: Oh, actually, we're not `export`ing it so I don't believe it's being passed to `dev-ui` or `dev-livechat`. It is being explicitly passed to `build-ui` and `build-livechat` which is to be expected.

joshbwlng:
On second thought, instead of spending time and effort on a halfway solution, I'd rather drop Makefiles entirely once Livepush-based development is more stable. Will start working on dropping Makefiles in a separate PR next week.

PR #7462, opened 2021-11-23
Fix exception in chart lens caused by missing contract type

LucianBuzzo:
The chart lens uses a in memory default contract for setting up the
initial chart configuration. This contract did not have a type set,
which caused a downstream error that made the lens crash.
A long term fix for this kind of problem is to add correct type
annotations to the UI, but I will leave a later PR.

Change-type: patch

***

test suite!


LucianBuzzo:
@balena-ci I self certify!

LucianBuzzo:
@balena-ci rebase

joshbwlng:
@balena-ci rebase

PR #7473, opened 2021-11-24
Add support for authentication tokens in the authentication middleware

edufschmidt:


joshbwlng:
@edufschmidt You'll need to tell VersionBot the change type here as well with a `Change-type: minor` commit footer. (Made a similar comment on your `jellyfish-action-library` PR). Also, to be clear, these changes don't necessarily depend on https://github.com/product-os/jellyfish-action-library/pull/1377 correct? Without both PRs merged and deployed authentication tokens won't work, but the system will still function as it currently does?

apps/server/lib/http/middlewares.ts
```diff
@@ -13,8 +18,10 @@ const logger = getLogger(__filename);
 export const attachMiddlewares = (
 	rootContext,
 	application,
-	_jellyfish,
-	options,
+	jellyfish: JellyfishKernel,
+	options: {
+		guestSessionId: string;
```
joshbwlng, line 23: Just curious, is this the only option ever passed to `attachMiddlewares()`?

apps/server/lib/http/facades/action.ts
```diff
@@ -35,7 +35,7 @@ export class ActionFacade {
 
 	async processAction(
 		context,
-		sessionToken,
+		sessionId,
```
joshbwlng, line 38: If we're going to change the variable name, I'd rather we go with `session` instead to match the rest of our code base. Do a quick grep under `@balena/jellyfish-core/lib` for example and you'll see what I mean.

joshbwlng:
You probably just haven't had time to get to it yet, but don't forget to add some tests to check the new functionality. :v:

edufschmidt:
> You'll need to tell VersionBot the change type here as well with a Change-type: minor commit footer. (Made a similar comment on your jellyfish-action-library PR).

Thanks for the heads up @joshbwlng. Will do that.

> Also, to be clear, these changes don't necessarily depend on product-os/jellyfish-action-library#1377 correct? Without both PRs merged and deployed authentication tokens won't work, but the system will still function as it currently does?

Yes, the changes are independent, and the system should still function normally either with one or the two PRs merged.

apps/server/lib/http/middlewares.ts
edufschmidt: Yes, the `options` argument is only used once within the `attachMiddlewares()` function to access the `guestSessionId` attribute.

apps/server/lib/http/facades/action.ts
edufschmidt: I thought `sessionId` would be more discriminative/easier to follow as we also have session objects sometimes, but you do have a good point. Will update that.

edufschmidt:
> You probably just haven't had time to get to it yet, but don't forget to add some tests to check the new functionality. ✌️

Yes, will do.

edufschmidt:
@joshbwlng I wrote a couple of integration tests, and would greatly appreciate it if you could share some feedback :)

apps/server/lib/http/facades/action.ts
edufschmidt: `sessionId` was replaced with `session`

apps/server/test/integration/auth.spec.ts
```diff
@@ -0,0 +1,51 @@
...
+		Authorization: `Bearer ${context.session.id}`,
+	});
+	expect(!response.error);
+	expect(response.data.slug).not.toMatch('user-guest');
+});
+
+test('server should accept requests containing the session token in the authorization header and the session id in the x-session-id header', async () => {
+	const { response } = await getSelfWithHeaders({
+		Authorization: `Bearer ${context.sessionToken}`,
+		'X-Session-ID': context.session.id,
```
joshbwlng, line 39: Might be overkill, but what happens when only `X-Session-ID` is provided in a request?

apps/server/test/integration/helpers.ts
```diff
@@ -62,13 +63,18 @@ export const before = async (context) => {
...
+
 	// Force login, even if we don't know the password
 	context.session = await context.sdk.card.create({
 		slug: `session-${userCard.slug}-integration-tests-${uuid()}`,
 		type: 'session',
 		version: '1.0.0',
 		data: {
 			actor: userCard.id,
+			token: {
+				authentication: await bcrypt.hash(context.sessionToken, 12),
```
joshbwlng, line 76: Do we need the generated UUID stored in `context.sessionToken` or can we just generate it here with something like `bcrypt.hash(uuid(), 12),`?

joshbwlng: By the way, we have a number of tests failing in this repo at the moment (unrelated to your work). I'm currently looking into them and should have them fixed by the time you want to merge this PR.

edufschmidt: We use `context.sessionToken` in one of our tests:

```
test('server should accept requests containing the session token in the authorization header and the session id in the x-session-id header', async () => {
	const { response } = await getSelfWithHeaders({
		Authorization: `Bearer ${context.sessionToken}`,
		'X-Session-ID': context.session.id,
	});
	...
});
```

Since login is being "forced", unless I'm missing something we wouldn't have access to the token otherwise.

apps/server/test/integration/auth.spec.ts
edufschmidt: As of now, when only `X-Session-ID` is provided, the auth middleware will look for the token in the `Authorization` header and find nothing. Therefore, it will not be able to authenticate and should use the guest session. I will add a test case to cover this scenario as well.

edufschmidt: @joshbwlng, done.

apps/server/lib/http/auth.ts
```diff
@@ -0,0 +1,70 @@
+import {
+	JellyfishKernel,
+	SessionContract,
+} from '@balena/jellyfish-types/build/core';
+import bcrypt from 'bcrypt';
+import _ from 'lodash';
+
+// TODO: Move this definition elsewhere, and inject it into the middleware.
+const validateSessionIdAndTokenFunc =
```
joshbwlng, line 9: We normally don't name functions with `Func` etc in the name. I've done similar things in other languages before, but TypeScript makes this unnecessary. :+1:

apps/server/lib/http/auth.ts
```diff
@@ -0,0 +1,70 @@
...
+	(context, jellyfish: JellyfishKernel) =>
+	async (session: string, sessionToken?: string): Promise<boolean> => {
+		return jellyfish
+			.getCardById<SessionContract>(context, jellyfish.sessions!.admin, session)
+			.then((result) => {
```
joshbwlng, line 14: I usually prefer to go with `await` when possible instead of `then` to make things a bit clearer. It's not too bad here since its only one level deep, but I'd still prefer to use `await` instead to not only match the rest of the code base, but also to make future edits that may require their own `await`/`then` easier.

edufschmidt: Yes, it makes sense. Replaced this with `await` as suggested.

apps/server/lib/http/auth.ts
edufschmidt: My goal here was to avoid confusion, since I use the higher-order function to create a validator which is later assigned to a const with a similar name. Anyways, the `Func` suffix was removed, and the functions renamed.

apps/server/lib/http/auth.ts
```diff
@@ -0,0 +1,75 @@
...
+	(jellyfish: JellyfishKernel, options: { guestSession: string }) =>
+	async (request, _response, next) => {
+		const sessionHeader = request.headers['x-session-id'];
+		const authorizationHeader = request.headers['authorization'];
+
+		const authorizationHeaderValue = _.last(_.split(authorizationHeader, ' '));
+
+		request.session = options.guestSession;
+
+		// TODO: take validation function injected via options rather than building them here.
```
joshbwlng, line 53: What would be the benefit of injecting the validation functions via options instead of defining them in this file? If we do end up keeping them in this file, it might be worth exporting them so you can write some simple unit tests for them. Much easier to catch things early with unit tests than wait for integration tests to fail, causing us to spend more time digging into the failure to find the cause.

apps/server/lib/http/auth.ts
```diff
@@ -0,0 +1,75 @@
...
+		const validateSessionId = sessionIdValidator(request.context, jellyfish);
+		const validateSessionIdAndToken = sessionIdAndTokenValidator(
+			request.context,
+			jellyfish,
+		);
+
+		if (sessionHeader) {
```
joshbwlng, line 60: We should probably check for `sessionHeader` etc earlier on so we don't unnecessarily query the DB with the validation functions.

apps/server/lib/http/auth.ts
```diff
@@ -0,0 +1,75 @@
...
+				session,
+			);
+			return bcrypt.compare(sessionToken, card!.data.token!.authentication);
+		} catch (e) {
+			return false;
+		}
+	};
+
+// TODO: Move this definition elsewhere, and inject it into the middleware.
+const sessionIdValidator =
```
joshbwlng, line 25: I don't think this function is necessary. `sessionIdAndTokenValidator()` also queries the database for the session contract. We could return `false` if the card doesn't exist at that point.

apps/server/lib/http/auth.ts
```diff
@@ -0,0 +1,75 @@
+import {
+	JellyfishKernel,
+	SessionContract,
+} from '@balena/jellyfish-types/build/core';
+import bcrypt from 'bcrypt';
+import _ from 'lodash';
+
+// TODO: Move this definition elsewhere, and inject it into the middleware.
+const sessionIdAndTokenValidator =
```
joshbwlng, line 9: Nit: Rename function to `isAuthenticationTokenValid` as it returns a boolean.

apps/server/lib/http/auth.ts
```diff
@@ -0,0 +1,75 @@
...
+	(context, jellyfish: JellyfishKernel) =>
+	async (session: string, sessionToken?: string): Promise<boolean> => {
```
joshbwlng, line 11:
```suggestion
async function isValidAuthenticationToken(context: Context, jellyfish: JellyfishKernel, session: string, sessionToken: string): Promise<boolean> {
```

apps/server/lib/http/auth.ts
```diff
@@ -0,0 +1,75 @@
+import {
+	JellyfishKernel,
+	SessionContract,
+} from '@balena/jellyfish-types/build/core';
```
joshbwlng, line 4:
```suggestion
import {
  Context,
	JellyfishKernel,
	SessionContract,
} from '@balena/jellyfish-types/build/core';
```

apps/server/lib/http/auth.ts
```diff
@@ -0,0 +1,75 @@
...
+			return bcrypt.compare(sessionToken, card!.data.token!.authentication);
+		} catch (e) {
+			return false;
+		}
+	};
+
+// TODO: Move this definition elsewhere, and inject it into the middleware.
+const sessionIdValidator =
+	(context, jellyfish: JellyfishKernel) =>
+	async (session: string): Promise<boolean> => {
```
joshbwlng, line 27:
```suggestion
async function isValidSession(context: Context, jellyfish: JellyfishKernel, session: string): Promise<boolean> {
```

apps/server/lib/http/auth.ts
```diff
@@ -0,0 +1,75 @@
...
+	async (request, _response, next) => {
+		const sessionHeader = request.headers['x-session-id'];
+		const authorizationHeader = request.headers['authorization'];
+
+		const authorizationHeaderValue = _.last(_.split(authorizationHeader, ' '));
+
+		request.session = options.guestSession;
+
+		// TODO: take validation function injected via options rather than building them here.
+		const validateSessionId = sessionIdValidator(request.context, jellyfish);
```
joshbwlng, line 54: Why not call the function when needed instead of first storing them into a variable and then calling that variable?

apps/server/lib/http/auth.ts
edufschmidt: In my opinion, the main benefits of injecting stuff is that it allows us to keep the code cleaner and more maintainable. In this specific scenario, it would result in an auth middleware that doesn't need to be aware of Jellyfish at all. It will simply define a clear interface (e.g. `validateToken`), and a higher-level component will provide it with a function matching that interface. It might also have some second-order benefits, especially if we want to "componentize" the project. Regarding unit tests, I don't see any donwsides. We will still be able to run unit tests for the middleware (possibly with mocking the validation functions) and for the validation functions independently, which I think is even more granular than what we have now.

apps/server/lib/http/auth.ts
edufschmidt: I believe this is related to the discussion about injecting vs defining stuff locally within this file. In the first scenario we would only receive the validation function (the one currently being assigned to the variable), whereas if we opt to have everything in this file we should definitely do what you suggested.

joshbwlng: @edufschmidt A part of me wants to just go the simpler route and define things locally, but I'm not against injection. However, if you want to go in that direction, I would suggest completing the implementation and not leaving TODOs about. From my experience, a TODO basically means that we'll probably get busy with other work, forget about it, and if we're lucky come back to it in a years time. 😂

apps/server/lib/http/auth.ts
joshbwlng: I understand the benefits of injection, but I tend to err on the side of not making something injectable unless I'm convinced that it will actually be useful and/or necessary to do so. One upside of defining things locally and not making them injectable is that the end result is that much simpler. I've worked on projects that were over-engineered with too many injection options that only added confusion without many real world benefits. But like I said in my other comment, if you're convinced this is a definite plus and will be useful/necessary then I'd like to see your implementation and I'll back you up. 👍

edufschmidt: I understand @joshbwlng. Although I definitely believe JF would benefit from such a change, I think this is something that we should discuss together and considering the whole codebase. That said, I will accept your suggestion so that we can move forward with this PR, and bring the topic for discussion in one of our arch calls.

apps/server/lib/http/auth.ts
edufschmidt: As mentioned in the other comment, I removed the TODOs and accepted your suggestion.

apps/server/lib/http/auth.ts
```diff
@@ -0,0 +1,81 @@
...
+			const sessionIsValid = await isValidToken(
+				request.context,
+				jellyfish,
+				credentials[0],
+				credentials[1],
+			);
+			if (sessionIsValid) {
+				request.session = credentials[0];
+			}
+		} else if (credentials.length === 1) {
```
joshbwlng, line 70: @edufschmidt Did some local testing and found that this path is taken even when an `Authorization` header isn't set, for example on the initial login when executing `action-create-session`. This results in an unnecessary call to the database that in turn results in a Postgres error:
```
   47 export const authMiddleware =
   48   (jellyfish: JellyfishKernel, options: { guestSession: string }) =>
>> 49   async (request, _response, next) => {
+  50     console.log('=== authMiddleware');
   51     const authorizationHeader = request.headers['authorization'];
   52 
   53     // Split 'Bearer ' and <session_id>.<auth_token>
   54     const authorizationHeaderValue = _.last(_.split(authorizationHeader, ' '));
   55 
   56     // Split session ID and auth token
   57     const credentials = _.split(authorizationHeaderValue, '.');
+  58     console.log('credentials:', credentials);
   59 
   60     request.session = options.guestSession;
   61 
   62     if (credentials.length === 2) {
   63       const sessionIsValid = await isValidToken(
   64         request.context,
   65         jellyfish,
   66         credentials[0],
   67         credentials[1],
   68       );
   69       if (sessionIsValid) {
   70         request.session = credentials[0];
   71       }
   72     } else if (credentials.length === 1) {
+  73       console.log('credentials[0]:', credentials[0]);
   74       const sessionIsValid = await isValidSession(
   75         request.context,
   76         jellyfish,
   77         credentials[0],
   78       );
   79       if (sessionIsValid) {
   80         request.session = credentials[0];
   81       }
   82     }
   83     return next();
   84   };
```

Logs:
```
[Logs]    [1/11/2022, 11:56:39 AM] [api] === authMiddleware
[Logs]    [1/11/2022, 11:56:39 AM] [api] credentials: [ '' ]
[Logs]    [1/11/2022, 11:56:39 AM] [api] credentials[0]: 
[Logs]    [1/11/2022, 11:56:39 AM] [postgres] 2022-01-11 02:56:39.129 UTC [309] ERROR:  invalid input syntax for type uuid: "" at character 280
[Logs]    [1/11/2022, 11:56:39 AM] [postgres] 2022-01-11 02:56:39.129 UTC [309] STATEMENT:  SELECT to_jsonb(cards) || jsonb_build_object('version', CONCAT_WS('+', CONCAT_WS('-', CONCAT_WS('.', cards.version_major, cards.version_minor, cards.version_patch), NULLIF(cards.version_prerelease, '') ), NULLIF(cards.version_build, '') )) AS payload
[Logs]    [1/11/2022, 11:56:39 AM] [postgres] 	FROM cards
[Logs]    [1/11/2022, 11:56:39 AM] [postgres] 	WHERE cards.id = ''
[Logs]    [1/11/2022, 11:56:39 AM] [postgres] 	LIMIT 1
```

edufschmidt: Thanks for pointing it out @joshbwlng. I patched the auth middleware to account for the case in which the authorization header is missing in the request. I also tested my changes locally, and the Postgres error is gone.

joshbwlng:
@balena-ci retest

joshbwlng:
@edufschmidt This test is currently failing: https://github.com/product-os/jellyfish/blob/master/test/e2e/server/security.spec.js#L22

Source: https://ci.product-os.io/builds/347480#L61e1e197:3710

edufschmidt:
> @edufschmidt This test is currently failing: https://github.com/product-os/jellyfish/blob/master/test/e2e/server/security.spec.js#L22
> 
> Source: https://ci.product-os.io/builds/347480#L61e1e197:3710

@joshbwlng , I've fixed the test, and it's now passing when I run it using the NUC. I'm now seeing another test failing, which seems unrelated to this PR:

```
  ✖ integrations › should take application/jose balena-api webhooks 
  ─

  integrations › should take application/jose balena-api webhooks

  test/e2e/server/integrations.spec.js:55

   54:                                               
   55:   test.is(result.code, 200)                   
   56:   const response = JSON.parse(result.response)

  Difference:

  - 401
  + 200

  › test/e2e/server/integrations.spec.js:55:7
```
Apparently this has something to do with the `options.sync.isValidEvent()` call in the hook API handler. Do you happen to have any idea why this is failing?

joshbwlng:
@edufschmidt

> Once I executed the e2e tests in a fresh terminal, the error went away, but the balena-api test is now being skipped. Is this expected?

The test is skipped when the key isn't set as an environment variable: https://github.com/product-os/jellyfish/blob/c7829b9ac2c9a6bb6307faa1468f082cf1b7596b/test/e2e/server/integrations.spec.js#L18

>  it seems that the privateKey was empty on the NUC (which I find confusing given that it is available as a secret)

The secrets have been added to the repo, but the work hasn't been done to pass these into containers. We currently use real compose for tests in CI and balena (supervisors rendition of compose) so we need to make sure neither of these break when passing secrets through. (We will be leaving balenaCI + real compose sometime this year, but that's a different topic.)

joshbwlng:
Good news is the `test/e2e/server/integrations.spec.js` tests are passing:
```
sut_1                    | [e2e-server]   ✔ integrations › should take application/jose balena-api webhooks (113ms)
sut_1                    | [e2e-server]   ✔ integrations › should not be able to post a GitHub event without a signature
sut_1                    | [e2e-server]   ✔ integrations › should take a GitHub event with a valid signature
sut_1                    | [e2e-server]   ✔ integrations › should not ignore a GitHub signature mismatch
sut_1                    | [e2e-server]   ✔ integrations › should not be able to post an Outreach event without a signature
sut_1                    | [e2e-server]   ✔ integrations › should take an Outreach event with a valid signature
sut_1                    | [e2e-server]   ✔ integrations › should not ignore an Outreach signature mismatch
sut_1                    | [e2e-server]   ✔ integrations › /api/v2/oauth should return a url given outreach
```

Bad news is I found this other test failing [here](https://ci.product-os.io/builds/365053#L61e80269:3577):
```
sut_1                    | [e2e-server]   ✖ sessions › should fail when querying an invalid session with an invalid session 
sut_1                    | [e2e-server]   ─
sut_1                    | [e2e-server] 
sut_1                    | [e2e-server]   sessions › should fail when querying an invalid session with an invalid session
sut_1                    | [e2e-server] 
sut_1                    | [e2e-server]   test/e2e/server/sessions.spec.js:165
sut_1                    | [e2e-server] 
sut_1                    | [e2e-server]    164:                                    
sut_1                    | [e2e-server]    165:   test.is(result.code, 400)        
sut_1                    | [e2e-server]    166:   test.deepEqual(result.response, {
sut_1                    | [e2e-server] 
sut_1                    | [e2e-server]   Difference:
sut_1                    | [e2e-server] 
sut_1                    | [e2e-server]   - 404
sut_1                    | [e2e-server]   + 400
sut_1                    | [e2e-server] 
sut_1                    | [e2e-server]   › test/e2e/server/sessions.spec.js:165:7
sut_1                    | [e2e-server] 
sut_1                    | [e2e-server]   ─
sut_1                    | [e2e-server] 
sut_1                    | [e2e-server]   `--fail-fast` is on. 3 test files were skipped.
sut_1                    | [e2e-server] 
sut_1                    | [e2e-server]   1 test failed
```

edufschmidt:
@balena-ci retest

edufschmidt:
@balena-ci retest

joshbwlng:
@balena-ci retest

joshbwlng (approved):

PR #7530, opened 2021-12-02
Update to default plugin 21.1.130

LucianBuzzo:
This change includes fixes to the user->contact trigger.

Change-type: patch

***

test suite!


LucianBuzzo:
@balena-ci I self-certify!

PR #7531, opened 2021-12-03
Remove local.env

joshbwlng:
Change-type: patch

***

Remove `local.env` and references to it. This was kind of useful sometimes, but was ultimately a failed experiment as it ended up just causing more confusion in regards to how to set and use environment variables during development and testing.


joshbwlng:
@balena-ci retest

joshbwlng:
@balena-ci I self certify!

PR #7538, opened 2021-12-04
Remove DATABASE from Makefile

joshbwlng:
Change-type: patch

***

Remove `DATABASE` from the root Makefile.


joshbwlng:
@balena-ci I self certify!

PR #7543, opened 2021-12-04
Drop POD_NAME from Makefile

joshbwlng:
Change-type: patch

***

- Drop `POD_NAME` from root Makefile
- Drop `TEST_INTEGRATION_SKIP` from `make push`


joshbwlng:
@balena-ci I self certify!

PR #7552, opened 2021-12-05
Enable failFast in ava options

joshbwlng:
Change-type: patch

***

The current `AVA_OPTS` environment variable in `Taskfile.yml` is not used, so setting `failFast` to `true` in ava options in `package.json` to apply to all ava executions. We do this with jest in lib repos and will be moving to TypeScript/jest in this repo as well at some point so may as well get on the same page test runner options wise.

### DIFF
--- a/Taskfile.yml
+++ b/Taskfile.yml
@@ -3,7 +3,6 @@ version: '3'
 output: 'prefixed'
 
 env:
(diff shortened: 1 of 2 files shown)


joshbwlng:
@balena-ci I self certify!

PR #7561, opened 2021-12-07
Remove Discourse vars from docker-compose.test.yml

joshbwlng:
Change-type: patch

***

Remove `INTEGRATION_DISCOURSE_*` environment variables from `docker-compose.test.yml` as they shouldn't be needed.


joshbwlng:
@balena-ci retest

joshbwlng:
@balena-ci I self certify!

PR #7574, opened 2021-12-08
Add compose:monitor script, update docs

joshbwlng:
Change-type: patch

***

- Add `npm run compose:monitor` script
- Update docs to now reference `npm run` instead of `make`


joshbwlng:
@balena-ci I self certify!

PR #7590, opened 2021-12-09
Remove more vars from test compose config

joshbwlng:
Change-type: patch

***

Remove more environment variables that have sane fallbacks from `docker-compose.test.yml`.


joshbwlng:
@balena-ci I self certify!

PR #7599, opened 2021-12-10
Stop skipping numerous e2e ui tests

joshbwlng:
Change-type: patch

***

Apparently we've been skipping a large number of E2E UI tests for a while because of some `.only`s. I believe this is a mistake and we should be running these tests.

#### `test/e2e/ui/index.spec.js`
- views: Should be able to save a new view
- chat-widget: A user can start a Jellyfish support thread from the chat widget
- outreach: A user should be able to connect their account to outreach
  - just missing a nocked endpoint?
- Chat: When having two chats side-by-side both should update with new messages
#### `test/e2e/ui/support.spec.js`
- My Participation shows only support threads that the logged-in user has participated in

### DIFF
--- a/test/e2e/ui/helpers.js
+++ b/test/e2e/ui/helpers.js
@@ -14,7 +14,8 @@ exports.puppeteerOptions = {
 
 		// Set extra flags so puppeteer runs on docker
 		'--no-sandbox',
-		'--disable-setuid-sandbox'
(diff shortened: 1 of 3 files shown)


joshbwlng:
@balena-ci rebase

joshbwlng:
Moved to https://github.com/product-os/jellyfish/pull/7862

PR #7614, opened 2021-12-11
Drop Redis vars from Makefile/compose

joshbwlng:
Change-type: patch

***

Drop Redis vars from Makefile and compose configs.


joshbwlng:
@balena-ci I self certify!

PR #7615, opened 2021-12-11
Drop registry vars from compose configs

joshbwlng:
Change-type: patch

***

Drop registry vars that have the same default values from the main compose config.


joshbwlng:
@balena-ci I self certify!

PR #7619, opened 2021-12-12
Move graphile-worker NO_LOG_SUCCESS var

joshbwlng:
Move the NO_LOG_SUCCESS environment variable used
by graphile worker. Without this variable being set
graphile worker will output all success logs, which
is unnecessary and noisy.

Change-type: patch


joshbwlng:
@balena-ci I self certify!

PR #7621, opened 2021-12-12
Move 'make deploy' to 'npm run push:lib'

joshbwlng:
Change-type: patch

***

Move the `make deploy` command to `npm run push:lib` and update docs to match. Also fix a bug found during testing that prevents packed lib tars from being copied.


joshbwlng:
@balena-ci I self certify!

PR #7622, opened 2021-12-12
Drop some port vars from root Makefile

joshbwlng:
Change-type: patch

***

Drop a number of unneeded port variables from the root Makefile.


joshbwlng:
@balena-ci retest

joshbwlng:
@balena-ci I self certify!

PR #7642, opened 2021-12-15
Update docs, add more helper npm scripts

joshbwlng:
Change-type: patch

***

- Add more helper npm scripts so we can execute everything we need from the project root
- Update local development and testing docs to reflect recent changes


joshbwlng:
@balena-ci I self certify!

PR #7643, opened 2021-12-15
Update frontend build config/scripts

joshbwlng:
Change-type: patch

***

- Clean up some old cruft leftover in `webpack.config.js`
- Use `webpack` for `apps/ui`'s `npm run build` (livechat is already doing this)


joshbwlng:
@balena-ci I self certify!

PR #7648, opened 2021-12-16
Fix mode in webpack build config

joshbwlng:
Fallback to production mode when building with webpack.
Explicitly set NODE_ENV for both prod and dev build commands.

Change-type: patch


joshbwlng:
@balena-ci I self certify!

PR #7658, opened 2021-12-17
Bump puppeteer to v13

joshbwlng:
Change-type: patch


joshbwlng:
@balena-ci I self certify!

PR #7679, opened 2021-12-20
Replace getActorKey in bootstrap.ts with import from jellyfish-worker

edufschmidt:
Change-type: minor

This PR replaces the `getActorKey` definition in `bootstrap.ts` with an import from `jellyfish-worker`, since they are exactly the same function. It also gets rid of the `SessionContractWithActor` interface, as the original `SessionContract` type defined in the `jellyfish-worker` module already contains an `actor` attribute.

### DIFF
--- a/apps/server/lib/bootstrap.ts
+++ b/apps/server/lib/bootstrap.ts
@@ -14,21 +14,13 @@ import type { core as coreType, JSONSchema } from '@balena/jellyfish-types';
 import { TriggeredActionContract } from '@balena/jellyfish-types/build/worker';
 import {
 	Contract,
-	SessionContract,
-	SessionData,
 	StreamChange,
 	TypeContract,
 } from '@balena/jellyfish-types/build/core';
 import { Transformer } from '@balena/jellyfish-worker/build/transformers';
+import { getActorKey } from '@balena/jellyfish-worker/build/utils';
 
 const logger = getLogger(__filename);
-
-// A session with a guaranteed actor set
-interface SessionContractWithActor extends SessionContract {
-	data: SessionContract['data'] & {
-		actor: string;
-	};
-}
 
 const SCHEMA_ACTIVE_TRIGGERS: JSONSchema = {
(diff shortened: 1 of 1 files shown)


apps/server/lib/bootstrap.ts
```diff
@@ -184,7 +139,7 @@ export const bootstrap = async (context, options) => {
 		try {
 			const key = await getActorKey(
 				context,
-				jellyfish,
+				jellyfish as any,
```
Ereski, line 142: We'd like to reduce in the long run `any`s instead of adding more. What would be required to give this the proper types?

Ereski (changes requested):

apps/server/lib/bootstrap.ts
```diff
@@ -14,22 +14,14 @@ import type { core as coreType, JSONSchema } from '@balena/jellyfish-types';
 import { TriggeredActionContract } from '@balena/jellyfish-types/build/worker';
 import {
 	Contract,
-	SessionContract,
-	SessionData,
 	StreamChange,
 	TypeContract,
 } from '@balena/jellyfish-types/build/core';
 import { Transformer } from '@balena/jellyfish-worker/build/transformers';
+import { getActorKey } from '@balena/jellyfish-worker/build/utils';
```
LucianBuzzo, line 21: These sorts of deep imports are dangerous, because they are not part of the top-level interface the worker module provides. This means that the internal implementation of the worker module is now tightly coupled to this code.
What can we do to eliminate the need for this at all?
FYI we discussed this "actor key" stuff a few months ago on an arch call and came to the conclusion that it's not needed (I jotted down some details [here](https://jel.ly.fish/improvement-jip-stop-generating-sessions-for-users-in-worker-code-adf47d50-62b7-43ef-91ef-349529d597a4) ).

joshbwlng:
Unnecessary as we are moving most of `apps/server/lib/` to the worker.

PR #7714, opened 2022-01-05
Improve webpack performance

karaxuna:
Change-type: minor

***

`source-map` is slowest both in terms of build and rebuild: https://webpack.js.org/configuration/devtool/ .
`eval` works well for build, excluding devtools field should have worked faster, but it does NOT for some reason. It decreased build time for me locally from 90s to 20s.

test suite!


karaxuna:
@balena-ci retest

taromurao:
I tested on my local machine. With this tweak, memory consumption stays constant. I can build JF also on a VM with "only" 4GB RAM.

taromurao (approved):

karaxuna:
@balena-ci rebase

PR #7742, opened 2022-01-10
workaround >1000 bootstrapped worker contracts

Hades32:
Due to Transformer types we now have lots of contracts which get loaded on startup. This is just a workaround to give as a bit more time, because we're already losing some contracts on startup, as we reached the limit of 1000 contracts per query result.

```
// count contracts (transformers, types, triggers)
for await (const r of [tf_schema,typpe_schema,trig_schema].map(q=>sdk.query(q) )) { console.log(r.length); }
 507
 236
 596

// count contracts (transformers, types, triggers) without draft versions
for await (const r of [tf_schema,typpe_schema,trig_schema].map(q=>sdk.query({...q, not:{properties:{version:{pattern:"-"}}}}) )) { console.log(r.length); }
 173
 138
 344
```

Load different contract types in separate but parallel queries to allow each to have its own limit.
Also, draft Transformers are filtered out anyway, so we don't query them

### DIFF
--- a/apps/server/lib/bootstrap.ts
+++ b/apps/server/lib/bootstrap.ts
@@ -13,7 +13,6 @@ import { getLogger } from '@balena/jellyfish-logger';
 import type { core as coreType, JSONSchema } from '@balena/jellyfish-types';
 import { TriggeredActionContract } from '@balena/jellyfish-types/build/worker';
 import {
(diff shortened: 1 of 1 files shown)


Hades32:
@balena-ci I self-certify!

PR #7774, opened 2022-01-17
Further restrict guest user access

LucianBuzzo:
Change-type: major
Signed-off-by: Lucian Buzzo <lucian.buzzo@gmail.com>

***

**Please remember to write tests for your changes**. We aim to automatically
deploy `master` to production, and we can't safely do this without a solid
test suite!

PR #7782, opened 2022-01-19
Use single purpose endpoints for actions performed by anonymous users

LucianBuzzo:
Using single purpose endpoints for actions performed by anonymous users
allow us to run the actions using a privileged user session and remove
the ability for the guest user to do it directly. This allows us to
reduce the scope of access that the guest user has, and therefore
anonymous users that the guest user represents.

Split out from https://github.com/product-os/jellyfish/pull/7774

Change-type: major

***

test suite!


LucianBuzzo:
@balena-ci I self-certify!

PR #7788, opened 2022-01-20
make Transformer log link dynamic

Hades32:
Now clicking on the log link should show you only logs from Transformers that ran due to the current commit


Hades32:
@cmfcruz do you know if there's a nicer way to create those Grafana/Loki links?

Hades32:
@balena-ci I self-certify!

PR #7792, opened 2022-01-21
Upgrade worker

karaxuna:
***

test suite!


apps/server/lib/http/routes.ts
```diff
@@ -172,7 +175,7 @@ export const attachRoutes = (
 
 				return response.status(200).json({
 					error: false,
-					data: _.omit(results.data, ['links']),
+					data: _.omit(results.data as any, ['links']),
```
Ereski, line 178: There's almost no valid reason for an `as any` cast. Is it really necessary here?

apps/server/lib/http/routes.ts
```diff
@@ -348,7 +351,7 @@ export const attachRoutes = (
 			}
 
 			return response.status(200).json({
-				access_token: createSessionResult.data.id,
+				access_token: (createSessionResult.data as any).id,
```
Ereski, line 354: Another `as any` cast

apps/server/package.json
```diff
@@ -25,7 +25,7 @@
   "dependencies": {
```
Ereski, line 25: Should also drop `jellyfish-plugin-base` from here

package.json
```diff
@@ -84,7 +84,7 @@
     "@balena/jellyfish-chat-widget": "^12.5.102",
```
Ereski, line 84: Drop `jellyfish-plugin-base` from here too

Ereski (commented):
@karaxuna Lots of small things slipped by

apps/server/lib/http/routes.ts
karaxuna: @Ereski `waitResults` should be generic and accept return type right?

apps/server/package.json
karaxuna: Dropped plugin-base, but there are some errors because of outdated plugin versions. I'll fix it when all plugins are updated.

apps/server/lib/http/routes.ts
Ereski: @karaxuna which `waitResults`?

apps/server/lib/http/facades/auth.ts
```diff
@@ -45,23 +47,23 @@ export class AuthFacade extends QueryFacade {
...
+			await this.queryAPI(
+				context,
+				session,
+				schema,
+				{
+					limit: 1,
+				},
+				ipAddress,
+			)
+		)[0] as Contract | null;
```
Ereski, line 60: `queryAPI` is an internal method. Why isn't it returning the right type?

apps/server/lib/http/facades/query.ts
```diff
@@ -57,12 +54,7 @@ export class QueryFacade {
 				schema,
 			});
 
-			const data = await this.jellyfish.query(
-				context,
```
Ereski, line 61: To avoid confusing anyone in the future, would be nice to rename everything of type `LogContext` to `logContext` or similar. `context` is too general for such a restricted value and JF has multiple `*Context` types. Not saying it has to be like that tho

apps/server/lib/http/routes.ts
karaxuna: @Ereski https://github.com/product-os/jellyfish/blob/master/apps/server/lib/http/routes.ts#L341

apps/server/lib/plugins.ts
```diff
@@ -11,24 +10,23 @@ import { GitHubPlugin } from '@balena/jellyfish-plugin-github';
 import { OutreachPlugin } from '@balena/jellyfish-plugin-outreach';
 import { ProductOsPlugin } from '@balena/jellyfish-plugin-product-os';
 import { TypeformPlugin } from '@balena/jellyfish-plugin-typeform';
+import { PluginManager } from '@balena/jellyfish-worker/build/plugin';
```
Ereski, line 13: No deep imports (`jellyfish-types` is an exception). `jellyfish-worker` already exports `PluginManager` from the root

Ereski (commented):
Based on the diffs it looks good. Just a few comments here and there

apps/server/lib/http/routes.ts
Ereski: How can you be sure that it is impossible for `createSessionResult.data` to be of any type other than object? https://github.com/product-os/jellyfish-types/blob/master/lib/queue/contracts/execute.ts#L16-L24

apps/server/lib/http/facades/auth.ts
karaxuna:
```typescript
const a: Contract[] = [];
const b = a[0];
```

Type of `b` is `Contract`, but I expected `Contract | undefined`. This is the reason why removing cast causes an error.

apps/server/lib/http/routes.ts
karaxuna: Not sure how it works internally. It may return value with one of these types, in which case we don't handle all of them properly and not only typing is a problem, otherwise it's just typing problem and we should tell `waitResults` what type we expect to be returned.

Ereski: @karaxuna sounds like this piece of code needs some extra error checking instead of a blanket `as any`

apps/server/lib/http/facades/auth.ts
Ereski: https://stackoverflow.com/questions/50647399/typescript-accessing-an-array-element-does-not-account-for-the-possibility-of-u

Ereski: :thinking:

karaxuna: @Ereski adding  `--noUncheckedIndexedAccess` sounds good for me, what do you think?

Ereski: Sounds good to me too! And in `jellyfish-config` if it's possible too!

karaxuna: @Ereski https://github.com/product-os/jellyfish-config/pull/60

Ereski:
There are some major version upgrades we gotta/can do. And remove the deprecated `request` package too

Ereski:
@joshbwlng @karaxuna in `server/lib/bootstrap.ts` we need to remove deep imports from the worker and move the `StreamChange` type into the core.

Ereski:
@karaxuna your force pushes completely erased the work I did yesterday

karaxuna:
> @karaxuna your force pushes completely erased the work I did yesterday

Oh my. What kind of changes are lost? :(

Ereski:
dependency updates, addition of missing deps, some types fixes and multiple import fixes. It was quite a bit

Ereski:
just need to know what you added to your last pushes to reconciliate our changes @karaxuna

LucianBuzzo:
Now failing with `Server error {"name":"TypeError","message":"jellyfish.defaults is not a function","stack":"TypeError: jellyfish.defaults is not a function\n    at loadCards (/usr/src/jellyfish/apps/server/build/card-loader.js:15:86)\n    at async bootstrap (/usr/src/jellyfish/apps/server/build/bootstrap.js:274:21)"}`

LucianBuzzo:
The issue with `.defaults` is resolved, now we need to fix some type errors that are thrown when unit testing the UI.

Ereski:
Tests failing with:

```
JellyfishDatabaseError: sorry, too many clients already
```

karaxuna:
@balena-ci retest

karaxuna:
@balena-ci retest

karaxuna:
@balena-ci retest

LucianBuzzo:
It's know failing to validate balena-api webhooks. I can't currently replicate the issue at home, so I've made a debug branch for balena-api to test this further https://github.com/product-os/jellyfish-plugin-balena-api/pull/306

karaxuna:
@LucianBuzzo I don't see failing tests [here](https://ci.product-os.io/builds/482357). I only see:

```
Could not find commit hash for previous e2e-server dump
```

LucianBuzzo:
@karaxuna urgh weird, that means the `jose` test is flakey then?

karaxuna:
@LucianBuzzo you mean the test that I fixed with last commit?

LucianBuzzo:
@karaxuna Amazing! I didn't even notice you'd pushed that 😂 - do the typings for that call need to be fixed?

LucianBuzzo:
Sick it passed! time to rebase and tighten up those types

LucianBuzzo:
@balena-ci retest

LucianBuzzo (approved):

PR #7814, opened 2022-01-25
UI: Improve typings for Interleaved component

LucianBuzzo:
This change improves the typings for the Interleaved lens, and tightens
up lens renderer types, although this hasn't been applied in all places.

This commit also introduces the pattern for typing redux connected
components, where three interfaces are defined:
- `StateProps`: The type for the props provided by `mapStateToProps()`
- `DispatchProps`: The type for the props provided by `mapDispatchToProps()`
- `OwnProps`: The type for the props provided by the parent component

These three types are combined for the code implementing the component:
```
type Props = StateProps & DispatchProps & OwnProps

interface State {
  internalComponentStateField: string
}

class MyComponent extends React.Component<Props, State> {
  ...
}
```

These three seperate types are then used as type paramaters to the
`connect` function, ensuring that the correct prop interface is exposed
to any parent component:

```
export default connect<StateProps, DispatchProps, OwnProps>
  (mapStateToProps, mapDispatchToProps)(MyComponent)
```

Change-type: patch

***

test suite!

### DIFF
--- a/apps/ui/lib/lens/full/View/Content.tsx
+++ b/apps/ui/lib/lens/full/View/Content.tsx
@@ -4,13 +4,12 @@ import { Box, Flex, Txt } from 'rendition';
 import { Icon } from '@balena/jellyfish-ui-components';
 import { core } from '@balena/jellyfish-types';
 import { ViewFooter } from '../../common/ViewFooter';
-import {
-	BoundActionCreators,
-	ChannelContract,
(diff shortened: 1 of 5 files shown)


LucianBuzzo:
@balena-ci retest

LucianBuzzo:
@balena-ci I self-certify!

PR #7910, opened 2022-02-14
Replace puppeteer with playwright

joshbwlng:
Change-type: patch

***

Replace `puppeteer` with `playwright` to reduce e2e test times and flakiness


joshbwlng:
@balena-ci I self certify!

PR #7917, opened 2022-02-14
Lazy load components

karaxuna:
Change-type: minor

***

Lazy load components that render conditionally, like routes, lenses, chat widget sidebar, login screen, widgets (some heavy libraries lazy loaded here)... This will split bundle into multiple pieces and fetch only those which are required.

This work is important for merging livechat into ui. We don't want to load entire JF for support users.


LucianBuzzo (approved):

karaxuna:
@balena-ci retest

PR #7933, opened 2022-02-15
UI: Stop using Redux to store view state

### LINKED ISSUE #7931: Don't store View lens state in redux store
[lucianbuzzo] Storing View contract data in the redux store generates unnecessary abstraction and indirection that leads to spaghetti code.

LucianBuzzo:
Fixes https://github.com/product-os/jellyfish/issues/7931

This change introduces `LiveCollection`, a component that wraps
together lens selection, filtering, sorting, pagination and live
updates.
Result state is stored inside this component and passed down to the
chosen lens renderer, instead of being stored globally in redux.
Seperating out this functionality as a seperate component also makes it
easier to re-use it elsewhere in the future (e.g. link tabs or inbox).

Change-type: patch

***

test suite!


apps/ui/lib/lens/full/SupportThread/index.tsx
```diff
@@ -21,13 +21,8 @@ const mapStateToProps = (state) => {
...
 	return {
-		actions: bindActionCreators(
-			pick(actionCreators, [
-				'addChannel',
-				'getActor',
-				'loadMoreViewData',
-				'removeChannel',
-			]),
+		actions: redux.bindActionCreators(
+			_.pick(actionCreators, ['addChannel', 'getActor', 'removeChannel']),
```
karaxuna, line 25: Not in the context of this pr, but why do we pick certain props from `actionCreators`, we could just:

```javascript
actions: redux.bindActionCreators(
  actionCreators,
  dispatch,
)
```

apps/ui/lib/components/CursorHOC.tsx
```diff
@@ -0,0 +1,141 @@
...
+interface Props {
+	query: Parameters<typeof sdk.getCursor>[0];
+	options: Parameters<typeof sdk.getCursor>[1];
+}
+
+interface State {
+	results: Contract[] | null;
+}
+
+export default class CursorHOC extends React.Component<Props, State> {
```
karaxuna, line 17: If you create `useCursor` hook, you can use it inside `CursorHOC`, this way you have both HOC and hook with minimal code.

karaxuna (commented):
I'm ok with this changes as it's main goal is to remove view state from redux.
For improving pagination, me and Carol had a conversation and I was experimenting around it in this [pr](https://github.com/product-os/jellyfish-chat-widget/pull/874) (See conversation link in pr).

apps/ui/lib/components/CursorHOC.tsx
karaxuna: Also, I think that this component is not HoC. See: https://reactjs.org/docs/higher-order-components.html

apps/ui/lib/lens/full/SupportThread/index.tsx
LucianBuzzo: I have no good reason for the `pick`, we should probably stop doing it just to make things simpler

apps/ui/lib/components/CursorHOC.tsx
LucianBuzzo: I think I'll refactor this PR to put the filters, lens selection and cursor inside a new component. This way it will be waaay easier to drop in the whole thing as a block anywhere we render a collection (like the link tabs)

LucianBuzzo:
@balena-ci retest

LucianBuzzo:
@balena-ci retest

LucianBuzzo:
@balena-ci I self-certify!

PR #7955, opened 2022-02-16
UI: Don't lazy load widgets

### LINKED ISSUE #7954: Rollback lazy loading of renderer widgets
[lucianbuzzo] It seems that rendition doesn't support lazy loading of widgets, resulting in markdown not being rendered in the UI. We need to rollback the lazy loading until we can figure out a fix.

LucianBuzzo:
Fixes https://github.com/product-os/jellyfish/issues/7954

Lazy loading widgets doesn't work correctly, resulting in markdown and
mermaid formats not being rendered correctly in the UI. This change
rolls back the lazyloading. We need to resolve this in rendition, by adding support for lazy loading widgets

Change-type: patch

***

test suite!


LucianBuzzo:
@balena-ci I self-certify!

PR #7957, opened 2022-02-16
Login as specific user

karaxuna:
Change-type: minor

***

User should be able to log in with different oauth providers like balena or outreach, we also can force to log in with specific user (need to embed livechat to dashboard, where we need to make sure that same users are logged in dashboard and in jellyfish). If jellyfish is loaded with `?login-as=user-karaxuna`, we make sure that current logged in user is `user-karaxuna`, otherwise we initiate oauth process with `balena` (hardcoded for now).


karaxuna:
@balena-ci rebase

LucianBuzzo (approved):

karaxuna:
@balena-ci rebase

karaxuna:
@balena-ci retest

karaxuna:
@balena-ci rebase

PR #7962, opened 2022-02-17
Fix clipped dropdown by preventing the LinkModal from scrolling in Y

edufschmidt:
This PR fixes the clipped dropdown issue in the link creation modal (`LinkModal`). This issue is due to the fact that `rendition` styles the modal with `overflow-y: auto`, but in this case we have a dropdown in it whose container is larger than the modal itself. By setting the `overflow-y` property to `inherit` the modal is no longer scrolling, and the dropdown is completely visible.

Change-type: patch

***

test suite!


LucianBuzzo:
@edufschmidt Can we fix this issue in rendition? Presumably the cloud dashboard and etcher face similar issues

LucianBuzzo (approved):

edufschmidt:
Yes @LucianBuzzo, in case it is indeed an issue and not a desired behavior. I suppose that here might be situations in which the modal is expected to be scrollable. What do you think?

PR #7965, opened 2022-02-17
Use LiveCollection component for segments

LucianBuzzo:
This change uses the generic LiveCollection component to display linked contracts, allowing filtering, search and lens functionality.

image

image


LucianBuzzo:
@balena-ci I self-certify!

PR #7971, opened 2022-02-18
Tidy up a number of UI issues around Segments/LiveCollection

LucianBuzzo:
See individual commits for details


LucianBuzzo:
@balena-ci I self-certify!

PR #7977, opened 2022-02-19
Fix oauth associate bug

karaxuna:
Change-type: patch

***

Associate function should return response in format `{ error, data }`.

test suite!


karaxuna:
@balena-ci I self certify!

PR #7978, opened 2022-02-19
Remove livechat service

karaxuna:
Change-type: major

***

This pr includes:
1. Move livechat tests to ui/livechat
2. Remove livechat service and related configurations
3. Redirect livechat.ly.fish to jel.ly.fish


LucianBuzzo (commented):
@karaxuna Looks good to me, 27k lines of code removed!
What do we need to do to ensure that `livechat.ly.fish` routes to the same UI as `jel.ly.fish`?

haproxy/haproxy.cfg
```diff
@@ -17,12 +17,9 @@ option forwardfor
 bind *:80
 http-request add-header X-Forwarded-Proto http
 
-acl host_ui_backend hdr_dom(host) -i jel.ly.fish.local
+acl host_ui_backend hdr_dom(host) -i jel.ly.fish.local livechat.ly.fish.local
```
karaxuna, line 20: @LucianBuzzo this change is for redirection. Tested locally.

karaxuna: @LucianBuzzo Not sure how this works. I assume there is a configuration for redirecting `*.local => *` on production.

LucianBuzzo: @karaxuna Have you checked in with @cmfcruz about what we need to do there?

joshbwlng:
@karaxuna Opened https://github.com/product-os/jellyfish/pull/8131 to remove livechat deployment/service from production and direct requests to `livechat.ly.fish` to go to ui pods instead. You might could merge that work into this PR, but it would probably make more sense to merge that one first and then remove livechat entirely with this one.

deploy-templates/jellyfish-product/product/config-manifest.yml
```diff
@@ -6,8 +6,6 @@ properties:
     type: string
   JELLYFISH_API_HOST:
     type: string
-  JELLYFISH_LIVECHAT_HOST:
-    type: string
```
joshbwlng, line 10: Rebased after merging https://github.com/product-os/jellyfish/pull/8131, but noticed `JELLYFISH_LIVECHAT_HOST` is being removed in this PR. The current livechat ingress (that still exists) depends on this variable. Are we OK with dropping `livechat.ly.fish` entirely or are there some other steps that need to take place beforehand so as to not interrupt support?

karaxuna: @joshbwlng we need to drop livechat entirely, we only need domain name to direct to ui service. Support chat already works with jel.ly.fish/livechat embedded in iframe. If I delete this configuration, will it prevent livechat.ly.fish redirection?

joshbwlng: @karaxuna We can drop all k8s configurations if we no longer need to provide anything at `livechat.ly.fish`. The current directing of requests from `livechat.ly.fish` to ui pods is because of the livechat ingress config and this variable. So, deleting the livechat ingress and/or this variable will prevent redirection. Does support chat not use the `livechat.ly.fish` domain?

karaxuna: @joshbwlng we should use livechat.ly.fish domain as iframe src in order to have different sessions on dashboard and on jellyfish. Currently both work with jel.ly.fish domain that causes shared localStorage. I'll leave k8s config untouched and just remove everything else.

joshbwlng (approved):

PR #7991, opened 2022-02-21
Fix error when vieweing support thread patterns

karaxuna:
Change-type: patch

***

tmp


karaxuna:
@balena-ci I self certify!

PR #8010, opened 2022-02-22
Update plugin default

karaxuna:
Change-type: minor

***

test suite!


karaxuna:
@balena-ci I self certify!

PR #8058, opened 2022-02-24
Update dependencies

Ereski:
Change-type: major

***

test suite!


Ereski:
Only one type error:

```
lib/bootstrap.ts:156:31 - error TS2345: Argument of type 'Redis' is not assignable to parameter of type 'CacheOptions'.
  Types of property 'socket' are incompatible.
    Type '{ host: string; port: string; }' is not assignable to type 'RedisSocketOptions | undefined'.
      Type '{ host: string; port: string; }' is not assignable to type 'RedisSocketCommonOptions & RedisTlsSocketOptions'.
        Property 'tls' is missing in type '{ host: string; port: string; }' but required in type 'RedisTlsSocketOptions'.
```

Weird, none of the other packages had issues

Ereski:
Not setting `tls` should be valid:

```ts
declare type RedisNetSocketOptions = Partial<net.SocketConnectOpts> & {
    tls?: false;
};
export interface RedisTlsSocketOptions extends tls.ConnectionOptions {
    tls: true;
}
export declare type RedisSocketOptions = RedisSocketCommonOptions & (RedisNetSocketOptions | RedisTlsSocketOptions);
```

Ereski:
Why is TS saying:

```
RedisSocketCommonOptions & RedisTlsSocketOptions
```

Instead of

```
RedisSocketCommonOptions & (RedisNetSocketOptions | RedisTlsSocketOptions)
```

Like in the declaration file?

Ereski:
So, the code as it is **should** work but TS is complaining because somehow it has the wrong definition for `RedisSocketOptions`.

Ereski:
Hmm, so manually setting tls to neither true nor false, nor forcing the call to accept the options as is makes it actually work. It fails to connect to redis

karaxuna:
@balena-ci retest

joshbwlng:
@balena-ci retest

joshbwlng:
@balena-ci retest

LucianBuzzo (approved):

PR #8099, opened 2022-03-01
Re-enable deployments

joshbwlng:
Reverts product-os/jellyfish#8069


joshbwlng:
@balena-ci I self certify!

PR #8105, opened 2022-03-01
Ignore status menu item if no access to user type

karaxuna:
Change-type: patch

***

Can't read user status options when opening top left menu if user does not have access to `user` type and ui breaks.


karaxuna:
@balena-ci rebase

karaxuna:
@balena-ci I self certify!

PR #8108, opened 2022-03-01
Remove SAMEORIGIN flag

karaxuna:
We are merging ui and livechat, so ui should be embedded inside iframe on dashboard. So, temporarly removing this restriction, until we set it properly to selected hosts.

Change-type: minor


karaxuna:
@balena-ci I self certify!

karaxuna:
@balena-ci rebase

karaxuna:
@balena-ci retest

PR #8109, opened 2022-03-01
Fix livechat timeline not visible

karaxuna:
Livechat container is not flex, this causes timeline to have 0 height.

Change-type: patch


karaxuna:
@balena-ci I self certify!

PR #8110, opened 2022-03-02
Accept JSON data on /oauth/callback

karaxuna:
This is required for transitioning from using `livechat.ly.fish` to `jel.ly.fish/livechat`. We first change JF oauth redirect url in balena api, so that when `livechat.ly.fish` is opened in iframe, it will redirect to balena oauth confirmation page, which redirects to `jel.ly.fish/oauth/callback` page with state in JSON format, because that was the state format used by livechat.

Change-type: minor


karaxuna:
@balena-ci I self certify!

PR #8115, opened 2022-03-02
Remove unnecessary oauth callback code

karaxuna:
This code was required for transitioning from livechat to ui without outage. Transition is complete, so removing extra code.

Change-type: minor


karaxuna:
@balena-ci I self certify!

karaxuna:
@balena-ci rebase

PR #8120, opened 2022-03-02
Do not show update confirmation popup if inside iframe

karaxuna:
Native `confirm` is used to ask for update confirmation, which is confusing for parent window users (balena users).

Change-type: minor

***

test suite!


LucianBuzzo (approved):

LucianBuzzo:
@karaxuna Is this resolved in prod or are we still having problems here?

PR #8131, opened 2022-03-04
Direct livechat.ly.fish requests to ui pods

joshbwlng:
Change-type: patch

***

Update deploy templates to reflect the recent work of serving livechat from the ui containers. This PR assumes that we no longer need the livechat containers at all and should instead direct all requests to `livechat.ly.fish` to the ui containers.


deploy-templates/jellyfish-product/product/deploy/kubernetes/generators/jellyfish-ui-ingress.tpl.yml
```diff
@@ -45,3 +45,14 @@ spec:
             name: jellyfish-ui
             port:
               number: 80
+  - host: '{{{JELLYFISH_LIVECHAT_HOST}}}'
```
joshbwlng, line 48: Ah, actually, we probably would instead want to keep the livechat ingress, but point it to the the ui pods

karaxuna (approved):
LGTM

karaxuna:
@balena-ci rebase

joshbwlng:
@karaxuna Thanks for the review, but I'd like to have @ab77 and/or @cmfcruz take a look before deploying. I haven't tested these configs and there is a chance that it could be botch deployment.

joshbwlng:
@balena-ci rebase

PR #8164, opened 2022-03-07
Remove service worker

karaxuna:
Remove old service worker code with unknown (and maybe unwanted) behavior.

Change-type: minor


joshbwlng:
@karaxuna Very nice! Are we thinking about bringing back a PWA implementation in the future or have we decided to drop that direction entirely? There are some benefits of PWA (installable as a psuedo-app for example), but I have always questioned how useful Jellyfish could be when offline (one of the main benefits of PWAs).

joshbwlng:
@balena-ci rebase

LucianBuzzo (approved):

karaxuna:
@balena-ci rebase

PR #8199, opened 2022-03-10
Move UI components

karaxuna:
Separate module for ui components is not needed any more, since we killed livechat service. And the code there is jellyfish specific.

Change-type: minor


LucianBuzzo (approved):

karaxuna:
@balena-ci rebase

PR #8200, opened 2022-03-10
UI: Add heuristic to improve performance of OmniSearch

LucianBuzzo:
This adds certain heuristics to OmniSearch to remove the potential for
the queries it generates to cause massive DB consumption on our RDS
based system. In experiments we found that the base enum used on the
virtual omnisearch view caused huge slowdown and stability issues, and
to a lesser extent the timeline search also cause problems.
With this change, omnisearch will no longer look for results in
a contracts timeline.
This is not the ideal solution, but will buy us time whilst we do
further investigation into why the query planning is so poor for this
request.
See http://localhost:8080/pattern-omnisearch-causes-massive-db-memory-consumption-8022ce1

Change-type: patch

***

test suite!


LucianBuzzo:
@balena-ci I self-certify!

PR #8208, opened 2022-03-11
UI: Optimize query for number of linked contracts

LucianBuzzo:
When loading relationship tabs, the UI would previously make a single
request for all linked contracts. This approach is very expensive and on
our production system would consistently time out for contract with more
than a handful of relationships. This change splits the query up into
a single request per relationship, which is much more performant and
significantly less memory consuming on the database.
This change also fixes an issue where the count would be incorrect if
the same link verb is used for multiple contract types.
Ideally we resolve this issue in the compiler, but for now this is
a sane patch.

Change-type: patch

***

test suite!


LucianBuzzo:
@balena-ci I self-certify!

PR #8209, opened 2022-03-11
UI: Fix bug where data doesn't reload when switching between tabs

LucianBuzzo:
Depends on https://github.com/product-os/jellyfish/pull/8208


LucianBuzzo:
@balena-ci I self-certify!

PR #8229, opened 2022-03-15
Fix localFS storage

joshbwlng:
Simplify and fix localFS storage implementation

Change-type: patch


joshbwlng:
@balena-ci I self certify!

LucianBuzzo:
@joshbwlng For future reference, a description of the bug being fixed here would be really useful 👍

PR #8239, opened 2022-03-16
UI: Disable full text search in timelines

LucianBuzzo:
When using the search functionality, the default behaviour is to additionally search within contract timeline for matching messages or whispers.
This additional search appears to be causing huge memory consumption on the production database, so for the sake of platform stability, I'm disabling the behaviour until we can fix the root cause of the issue.

Change-type: patch

***

test suite!


LucianBuzzo:
@balena-ci I self-certify!

PR #8244, opened 2022-03-16
Integrate with ory hydra

karaxuna:
This pr makes jellyfish an oauth provider using ory hydra. We should have same oauth flow when open support chat widget as we have on dashboard. User should login with jellyfish into livechat on jellyfish and login with balena on balena dashboard.


apps/server/lib/bootstrap.ts
```diff
@@ -91,10 +186,10 @@ export const bootstrap = async (logContext: LogContext, options: any) => {
 
 	// Create and initialize the worker instance. This will process jobs from the queue.
 	const worker = new Worker(
-		kernel,
+		kernel as any,
```
karaxuna, line 189: Type error because of using draft autumndb version.

apps/server/lib/bootstrap.ts
```diff
@@ -302,6 +397,73 @@ export const bootstrap = async (logContext: LogContext, options: any) => {
...
+			from: {
+				id: jellyfishOauthProvider!.id,
+				type: jellyfishOauthProvider!.type,
+			},
+			to: {
+				id: jellyfishOauthClient.id,
+				type: jellyfishOauthClient.type,
+			},
+		},
+	});
```
karaxuna, line 451: Need to preinsert and link not only jellyfish oauth contracts, but also outreach and balena api.

docker-compose.yml
```diff
@@ -267,7 +280,39 @@ services:
...
+      - hydra-migrate
+      - postgres
+      - mdns
+    environment:
+      - URLS_SELF_ISSUER=http://hydra.ly.fish.local
+      - URLS_CONSENT=http://jel.ly.fish.local/oauthprovider/consent
+      - URLS_LOGIN=http://jel.ly.fish.local/oauthprovider/login
+      - URLS_LOGOUT=http://jel.ly.fish.local/oauthprovider/logout
+      - DSN=postgres://docker:docker@postgres:5432/jellyfish?sslmode=disable&max_conns=20&max_idle_conns=4
+      - SECRETS_SYSTEM=youReallyNeedToChangeThis
```
karaxuna, line 302: Not sure about this, see: https://www.ory.sh/docs/hydra/guides/secrets-key-rotation#rotation-of-hmac-token-signing-and-database-and-cookie-encryption-keys

apps/server/lib/bootstrap.ts
```diff
@@ -302,6 +397,73 @@ export const bootstrap = async (logContext: LogContext, options: any) => {
 		}),
 	);
 
+	// Manually sync oauth clients
+	// TODO: This should ideally be completely automated, but this would
```
LucianBuzzo, line 401: I think we should probably have a sensible load priority for default contracts. Off the top of my head it should be:
1. types
2. relationships
3. actions
4. triggered actions
5. everything else

karaxuna: @LucianBuzzo there is a problem with preinserting links, because you don't know ids of contracts which you are linking before they are inserted.

docker-compose.yml
```diff
@@ -6,7 +6,10 @@ volumes:
 
 services:
   postgres:
-    image: balena/open-balena-db:v5.1.0
+    image: balena/jellyfish-postgres
+    build:
+      context: ./apps/postgres
+      dockerfile: Dockerfile
```
karaxuna, line 12: Created separate database for hydra.

apps/haproxy/haproxy.cfg
```diff
@@ -190,4 +194,4 @@ backend ocsp-backend
 		server cfssl-ocsp balena-ca:8889 resolvers docker-bridge-resolver resolve-prefer ipv4 check port 8889
 
 backend wifi-connect-backend
-		server wifi-connect 192.168.42.1:1337 check port 1337
+		server hydra hydra:4444 resolvers docker-bridge-resolver resolve-prefer ipv4 check port 4444
```
joshbwlng, line 197: Probably shouldn't be replacing `wifi-connect` with `hydra` here?

apps/server/lib/plugins/jellyfish.ts
```diff
@@ -0,0 +1,180 @@
...
+import { defaultEnvironment as environment } from '@balena/jellyfish-environment';
+import { hydraAdmin } from '../hydra-sdk';
+
+export const jellyfishPlugin = ({
+	kernel,
+}: {
+	kernel: Kernel;
+}): PluginDefinition => {
+	return {
+		slug: 'plugin-jellyfish',
```
joshbwlng, line 14: I really don't like the idea of (1) naming a plugin for Jellyfish `plugin-jellyfish` and (2) defining it inline in the server. This looks like it shoule belong in a `jellyfish-plugin-oauth` repo or something instead.

apps/server/test/integration/oauth.spec.ts
```diff
@@ -351,7 +350,7 @@ balenaApiTest(
 
 		nock.cleanAll();
 
-		nock(environment.integration['balena-api'].oauthBaseUrl)
+		await nock('https://dashboard.balena-cloud.com')
```
joshbwlng, line 353: Do we really need the `await`s for nock?

package.json
```diff
@@ -42,9 +42,7 @@
     "dev:ui": "(cd apps/ui && npm run dev)"
   },
   "deplint": {
-    "files": [
-      "apps/server/lib/**/*.ts"
-    ],
+    "files": [],
```
joshbwlng, line 45: We shouldn't be skipping lint here, I'd be interested in what you are trying to avoid by removing this.

apps/postgres/create-hydra-db.sh
```diff
@@ -0,0 +1,5 @@
+#!/bin/bash
+
+psql --username docker <<-EOSQL
+	CREATE DATABASE hydra;
+EOSQL
```
joshbwlng, line 5: Is this something we could do in `apps/hydra/start.sh` instead?

package.json
karaxuna: @joshbwlng see https://jel.ly.fish/thread-70e3970b-c722-4341-989f-a74278835e92

joshbwlng: @karaxuna Right, I think it would be better to change the way the code is organized instead of disabling lint. One idea is to move `plugin-jellyfish` to be an actual plugin in its own repo like `jellyfish-plugin-oauth`. You may end up with the `hydraAdmin()` code in two places, but I think it would be a simpler approach that better matches how we manage other plugins.

joshbwlng: I can help you set up a `jellyfish-plugin-oauth` repo, might go ahead and start that today if I have the time.

joshbwlng: Created the repo and started bringing in your work here: https://github.com/product-os/jellyfish-plugin-oauth/pull/1
Need to fix the two todos I've commented on in the integration code, but I think everything else should be good once we solve that.

apps/postgres/create-hydra-db.sh
joshbwlng: The reason I ask is that I'd like to keep things as isolated as possible. I think it should be possible if we install `psql` in the `hydra` container?

apps/server/lib/plugins/jellyfish.ts
joshbwlng: As I commented below, I've opened up a `product-os/jellyfish-plugin-oauth` repo and started bringing over your code: https://github.com/product-os/jellyfish-plugin-oauth/pull/1

karaxuna:
@balena-ci retest

.github/workflows/balena.yml
```diff
@@ -32,8 +32,9 @@ env:
     export-database \
     import-database \
     e2e-server-previous-dump
-
-  LIVECHAT_PORT: 443
+  INTEGRATION_JELLYFISH_APP_ID: jellyfish
+  INTEGRATION_JELLYFISH_APP_SECRET: changeme
+  HYDRA_ADMIN_HOST: hydra:4445
```
joshbwlng, line 37: @karaxuna Would this be better as `http://hydra:4445`? Found that `nock` throws an error when you only set `hydra:4445`, seemingly requiring the protocol be provided. Currently covering this in `integration-server` tests by setting `HYDRA_ADMIN_HOST=http://hydra:4445` in `apps/server/package.json` to override the value given here.

karaxuna: I still got same error. Can't understand why request does not get mocked.

karaxuna:
@balena-ci retest

.github/workflows/balena.yml
joshbwlng: The `integration-server` tests are passing, now its getting stuck at `e2e-ui` when attempting to log in as the test user with the sdk.

karaxuna:
@balena-ci rebase

.github/workflows/balena.yml
```diff
@@ -32,8 +32,9 @@ env:
     export-database \
     import-database \
     e2e-server-previous-dump
-
-  LIVECHAT_PORT: 443
+  INTEGRATION_JELLYFISH_APP_ID: jellyfish
+  INTEGRATION_JELLYFISH_APP_SECRET: changeme
```
joshbwlng, line 36: Can we drop these? I see we have defaults in `jellyfish-environment` set here: https://github.com/product-os/jellyfish-environment/blob/master/lib/variables/integration/defaults.ts#L18-L19

.github/workflows/balena.yml
```diff
@@ -32,8 +32,9 @@ env:
...
+  HYDRA_ADMIN_HOST: http://hydra:4445
```
joshbwlng, line 37: Maybe we could also drop this line? It seems to be the same value as the default defined here: https://github.com/product-os/jellyfish-environment/blob/master/lib/variables/hydra/defaults.ts#L1

apps/server/lib/http/routes.ts
```diff
@@ -7,6 +7,7 @@ import type {
 	Sync,
 	Worker,
 } from '@balena/jellyfish-worker';
+import { hydraAdmin } from '@balena/jellyfish-plugin-oauth/build/actions/action-sync-oauth-client';
```
joshbwlng, line 10: We shouldn't deep import from under `build/` if possible. I've updated `plugin-oauth` to export this function from the main index file so you should be able to `import { hydraAdmin } from '@balena/jellyfish-plugin-oauth';`

https://github.com/product-os/jellyfish-plugin-oauth/pull/6

karaxuna: Thanks Josh!

apps/server/run.sh
```diff
@@ -27,8 +30,8 @@ INTEGRATION_BALENA_API_PUBLIC_KEY_PRODUCTION=${INTEGRATION_BALENA_API_PUBLIC_KEY
 cd /usr/src/jellyfish/apps/server
 
 if [[ $BALENA_APP_NAME == 'localapp' ]] && [[ $BALENA_APP_ID -eq 1 ]]; then
-		npm run dev
+		NODE_ENV=development ./node_modules/.bin/nodemon --inspect=0.0.0.0 ./lib/index.ts
 else
 		# (TBC) should then become a systemd service (e.g jellyfish-api)
-		npm start
```
joshbwlng, line 33: I'm pretty sure this breaks the local development (non-Livepush) flow...

joshbwlng, line 33: I'm pretty sure this breaks the local development (non-Livepush) flow...

apps/server/package.json
```diff
@@ -18,27 +18,27 @@
     "test": "catch-uncommitted --skip-node-versionbot-changes && npm run lint && npm run test:integration",
     "doc": "typedoc lib/",
     "prepack": "npm run build",
-    "start": "node --abort-on-uncaught-exception --stack-trace-limit=100 build/index.js",
-    "dev": "NODE_ENV=development nodemon --inspect=0.0.0.0 ./lib/index.ts"
```
joshbwlng, line 22: @karaxuna I'd like to keep `npm run dev`. Could we keep `npm start` and `npm run dev` as-is and call them from `run.sh`? My reason is keeping the ablity to run services locally (not in balena Livepush, but on local PC directly).

PR #8256, opened 2022-03-18
Improve pagination indicators on support thread lens

LucianBuzzo:
- Adds a `+` icon if there are more results available
- Only shows spinner if there are more results available


LucianBuzzo:
@balena-ci I self-certify!

PR #8278, opened 2022-03-22
UI: Refactor how "slices" work

LucianBuzzo:
This change makes slices simply a set of JSON schemas that can be used to pre-filter queries.
This change dramatically simplifies how this feature works and removes a bunch of intermediate data formats.
As part of this change, the whole set of user defined filters is now preserved on reload, which is a nice change.


LucianBuzzo:
@balena-ci I self-certify!

PR #6847, opened 2021-07-06
Develop and run Jellyfish on balenaOS in balenaCloud

ab77:
Quickstart: https://github.com/product-os/jellyfish/blob/ab77/jellyfish-on-balena/docs/developing/on-balena.markdown

- https://github.com/product-os/environment-production/pull/277
- https://github.com/product-os/secrets/pull/46

TBCs:
- simplify/refactor all docs (i.e. remove docker|compose references, etc.) for development on-balena only
- refactor defaults handling between Taskfiles(s), balena.yml, package.json, etc. and jellyfish-environment
- refactor apps into systemd services and confd runtime (support mDNS .local resolution)
- rewire Grafana/Prometheus metrics/logs

Operational:
- remove docker test from branch protection rules on repository
- scale up `jellyfish-compute-resources-AutoScalingGroupSpot-F3K16JXB3ZTM` AutoScaling group via CFN (min x2)
- (optionally) update AWS/EC2 instance size
- update DNS to point to new ALBs (IPv4/IPv6 via AWS CNAME)

```
{api,jel,livechat}.ly.fish => jelly-LoadB-TAFQRKGBN0E5-167937939.us-east-1.elb.amazonaws.com
```

- (optionally) scale down Kubernetes productOS nodegroup(s) to x2 (minimum)
- remove Jellyfish Kubernetes assets from AWS/EKS production cluster (production-eks)
- remove balenaCI (Concourse) Jellyfish pipelines/assets

Future:
- refactor/replace Postgres/S3 tests with DB migrations (remove test dependency on external S3 object store in staging AWS account)

Change-type: major

PR #8284, opened 2022-03-23
Make grafana/prometheus first class apps

joshbwlng:
Change-type: patch

***

Move `grafana`, `prometheus`, and `haproxy` directories to live under `/apps`. Also update `docker-compose.yml` to match, making `grafana` and `prometheus` first class apps and not storing them away in a separate `docker-compose.monitor.yml` file. 

The only downside is the grafana dashboard config being a duplicate of what we already have under `deploy-templates`. This won't be a problem for long though as we will be dropping `deploy-templates` completely when we make the switch to running on balena.

Part of the work to slowly merge in parts of https://github.com/product-os/jellyfish/pull/6847

ToDo:
- [x] Grafana dashboards aren't showing up as expected, looking into it
- [x] Would also like to route 80->9090 and 80->3000 for prometheus and grafana respectively
- [x] Test on balena device with livepush


joshbwlng:
@balena-ci rebase

joshbwlng:
@balena-ci I self certify!

PR #8310, opened 2022-03-25
Update @balena/jellyfish-worker to 21.0.0

joshbwlng:
Change-type: patch

***

Update `@balena/jellyfish-worker` to 21.0.0, dropping `@balena/jellyfish-queue` as it is now merged into the worker. Also bumping all plugins after bringing these latest changes into them as well.


joshbwlng:
@balena-ci I self certify!

PR #8332, opened 2022-03-27
Set grafana log level to error

joshbwlng:
Grafana logs were a bit too noisy

Change-type: patch


joshbwlng:
@balena-ci I self certify!

PR #8360, opened 2022-03-31
Lucianbuzzo/tab selection persist

LucianBuzzo:
This PR converts many Lenses to use the new `TabbedContractLayout` component, and also adds preservation of tab selection and relationship selection to the UI. The selection is per type, so navigating to a different contract of the same type will keep the same tabs active.


LucianBuzzo:
@balena-ci I self-certify!

PR #8363, opened 2022-03-31
UI: Use HTTP API for initial contract data load

LucianBuzzo:
Using the HTTP API for the initial data load is notably faster than
waiting for a websocket connection and querying the payload over WSS,
resulting in a snappier UX.

Change-type: patch

***

test suite!


LucianBuzzo:
@balena-ci I self-certify!

PR #8364, opened 2022-04-01
Increase timeouts on haproxy pg and redis to prevent timeouts from JF

ramirogm:
Change-type: patch

***

test suite!


apps/haproxy/haproxy.cfg
```diff
@@ -33,14 +33,18 @@ acl host_prometheus_backend hdr_dom(host) -i prometheus.ly.fish.local
 use_backend prometheus_backend if host_prometheus_backend
 
 frontend tcp_5432_in
-mode tcp
-bind *:5432
-default_backend postgres_backend
+  bind *:5432
+  mode tcp
+  log global
```
joshbwlng, line 38: Do we need to change log settings? If I'm not mistaken this change isn't really useful as we don't have anything log-related configured in the `global` section at the top of the file.

apps/haproxy/haproxy.cfg
```diff
@@ -33,14 +33,18 @@ acl host_prometheus_backend hdr_dom(host) -i prometheus.ly.fish.local
...
+  timeout client 1440m
```
joshbwlng, line 39: I would rather set timeout settings in a global way and not on a per-service basis. This is done in this file in the `defaults` section, and in a slightly more legible way here: https://github.com/product-os/jellyfish/pull/6847/files#diff-ae1fba26d587e6e571217dbbd34ecdaca8b9adc59423ef18c144193debf30995

Maybe we can bring timeout related settings from that PR into this one and tweak the timeouts to be a bit more if necessary. Also, is there a reason you went with 1440 minutes? It seems a bit excessive. There are downsides to turning this up too much: https://discourse.haproxy.org/t/turn-off-all-timeouts/91

joshbwlng: We need to remember that haproxy will be used in production before too long and not assume that these settings are only for development.

joshbwlng:
@ramirogm Thanks for the PR! Left a quick comment, but still need to confirm functionality when working with Jellyfish both locally and through Livepush. Just a note that this haproxy service is currently only used in testing/development, but will be used in production before too long so we should keep that in mind when making any changes to its configuration.

Also, when all services are running in compose/balena, connections from the JF server to the db/cache are direct and not through haproxy so no timeouts like the ones you're currently facing should occur. Nothing wrong with your PR in this regard, just wanted to make sure we're on the same page. 👍

apps/haproxy/haproxy.cfg
```diff
@@ -33,14 +33,18 @@ acl host_prometheus_backend hdr_dom(host) -i prometheus.ly.fish.local
...
+  default_backend postgres_backend
 
 frontend tcp_6379_in
-mode tcp
-bind *:6379
-default_backend redis_backend
+  bind *:6379
+  mode tcp
+  log global
+  timeout client 1440m
```
joshbwlng, line 46: Any reason you went with 1440m? I think you can set it to 0 for infinite. That being said, we will probably want to remove the ability to connect to postgres/redis through haproxy when haproxy is used in production. Not sure about this yet, just thinking out loud. 🤔

apps/haproxy/haproxy.cfg
ramirogm: Correct, this is mostly a leftover, I'll remove it

apps/haproxy/haproxy.cfg
ramirogm: 1440m is just a big number not to experience the issue. More about this below

ramirogm:
Thanks for the feedback. I didn't realize the haproxy config was meant to also be used in prod later, but that makes sense. And we should be careful about making changes

>Also, when all services are running in compose/balena, connections from the JF server to the db/cache are direct and not through haproxy so no timeouts like the ones you're currently facing should occur
Right

As a general comment, this PR is a workaround for an issue that I think could also occur in other cases. The `haproxy` configuration causes the timeout after 60secs of inactivity. We can increase that timeout, ok. But the issue that bothers me is that JF doesn't handle the error and node exits. It should reconnect IMO. I still need to track down why this is happening as part of my onboarding to JF.

joshbwlng:
@ramirogm I completely agree that JF should intelligently reconnect instead of erroring out. We should still push forward with this PR though and get the haproxy config in as good of shape as possible. I'm off today, so will try and come back to this tomorrow. :+1:

joshbwlng:
@ramirogm Breaking out the readme changes you made in this PR to https://github.com/product-os/jellyfish/pull/8372

joshbwlng:
Also breaking out the haproxy timeout settings into this PR: https://github.com/product-os/jellyfish/pull/8374
Should note that I've dropped the `retries 2` changes as the default for `retries` is 3 and I don't see any reason why we would want to manually drop that down to 2. [Source](https://www.haproxy.com/blog/haproxy-layer-7-retries-and-chaos-engineering/)

I've confirmed locally that the timeout we were seeing with that particular local development setup no longer occurs. Also confirmed that nothing is changed when running Jellyfish through Livepush on a balena device. Let me know though if you run into anything or have any issues with my PR.

ramirogm:
Thanks @joshbwlng, I'll proceed with the `docs/developing` updates in a new PR

PR #8365, opened 2022-04-01
Update push-lib to work with new apps

joshbwlng:
Just adding a file exists check for
package-lock.json instead of assuming it
exists as some apps are not node-based.

Change-type: patch

***

test suite!


joshbwlng:
@balena-ci I self certify!

PR #8369, opened 2022-04-04
Use cursors for loading timeline data

LucianBuzzo:
This change switches from using a centralized stream for timeline data,
to a separate cursor, held in memory in the timeline component. The
cursor allows for smarter pagination and a simpler data flow.
This change also fixes an issue where timeline stream would not
disconnect when the timeline component unmounted and fixes another issue
where the timeline would paginate multiple times on initial load.


LucianBuzzo:
@balena-ci I self-certify!

PR #8374, opened 2022-04-05
Update haproxy config

joshbwlng:
Update whitespace to match official examples.
Update timeouts to help avoid unnecessary disconnects.

Change-type: patch

***

Main change here is in the haproxy timeout settings. When using `npm run compose:database` to run Postgres and Redis through haproxy in local compose to work with a local instance of the Jellyfish server not in compose, the connection would timeout and cause the server to crash. This crash is something that should be dealt with in AutumnDB (more intelligent reconnect attempts when necessary), but we also don't need to be strict at all with Postgres and Redis connection timeouts as this is only used during local development. This will change though when we move to run production Jellyfish on balena and therefore use haproxy in production, in which case we should probably drop Postgres and Redis haproxy entries entirely.

Original PR: https://github.com/product-os/jellyfish/pull/8364

### DIFF
--- a/apps/haproxy/haproxy.cfg
+++ b/apps/haproxy/haproxy.cfg
@@ -1,10 +1,10 @@
 global
-tune.ssl.default-dh-param 1024
+  tune.ssl.default-dh-param 1024
 
 defaults
-  timeout connect 5000
-  timeout client 60000
-  timeout server 60000
+  timeout connect 5s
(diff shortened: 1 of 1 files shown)


joshbwlng:
@balena-ci I self certify!

joshbwlng:
@balena-ci rebase

PR #8377, opened 2022-04-05
UI: Don't use slice filtering when viewing data in a tab

LucianBuzzo:
Previously, when viewing linked data, slices would be used (e.g.  prefiltering for status:open). However in most cases, this is not appropriate and can be misleading.
With this change, slices will only be used when rendering results from a View.

Change-type: minor

***

test suite!


LucianBuzzo:
@balena-ci I self-certify!

ghost (approved):

PR #8381, opened 2022-04-05
UI: Load all results when generating a CSV download

LucianBuzzo:
Previously, only the results loaded into the client would be downloaded.
Now all matching results are loaded, paginating data if necessary.

Change-type: patch

***

test suite!


Jazzagi:
@LucianBuzzo Can you give me a ping when this is merged please? I'd like to download forums data but my index finger won't thank me if i try without this improvement! :)

LucianBuzzo:
@balena-ci I self-certify!

PR #8420, opened 2022-04-10
Set NODE_ENV=development for server dev

joshbwlng:
Change-type: patch

***

Ensure that `NODE_ENV` is set to `development` when running the server during development.


joshbwlng:
@balena-ci I self certify!

PR #8447, opened 2022-04-14
Fix switching between lenses

karaxuna:
`/all` view does not have `data.slices` field, causing `slice` being null.
Does kanban lens make sense if there is no slice? This is a quick fix, need to discuss this further.

Change-type: patch

### DIFF
--- a/apps/ui/lib/lens/common/LiveCollection/Renderer.tsx
+++ b/apps/ui/lib/lens/common/LiveCollection/Renderer.tsx
@@ -79,7 +79,7 @@ const getTypesFromSchema = (schema) => {
 
 const setSliceFilter = (filters, lens, slice) => {
 	// We only want to filter by slice if the lens does not supports slices itself!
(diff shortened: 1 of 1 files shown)


LucianBuzzo (approved):

LucianBuzzo:
@karaxuna Is there a sane default we could use in the Kanban lens if there are no slices? I can't think of one...
It might be the case that Kanban is only offered if it can actually divide up the results with a slice...

LucianBuzzo:
@karaxuna Yeah that could work, but it could also be horrendous 😂 
Maybe the sane thing is just to not render the Kanban lens if there are no slices. I'm not sure if we currently have a mechanism for doing this though

karaxuna:
> Yeah that could work, but it could also be horrendous

Agree. 

> Maybe the sane thing is just to not render the Kanban lens if there are no slices. I'm not sure if we currently have a mechanism for doing this though

I could not find a quick way to hide Kanban lens.

karaxuna:
@balena-ci rebase

PR #8468, opened 2022-04-19
Update registry/mdns services

joshbwlng:
Simplify registry and mdns service configuration,
resolving neverending service restarts

Change-type: patch


joshbwlng:
@balena-ci I self certify!

PR #8493, opened 2022-04-25
Improve timeline pagination logic

LucianBuzzo:
This change fixes a number of issues with timeline pagination.
1) It fixes an issue where the timeline would always paginate multiple times on
	 initial load.
2) It fixes an issue where the timeline would "stick" at the top and
	 continue to load every page.
3) It fixes an issue where, upon paginating, the timeline would jump,
	 not retaining its position.

Change-type: patch

***

test suite!


LucianBuzzo:
@balena-ci I self-certify!

PR #8499, opened 2022-04-26
Update to worker v23

joshbwlng:
Change-type: patch

***

test suite!


apps/server/lib/http/facades/action.ts
```diff
@@ -22,14 +27,14 @@ interface ActionFacadeOptions {
 	files?: FileDetails[];
 }
 
-type ActionPayload = Parameters<Producer['enqueue']>[2];
+type PreRequest = Parameters<Worker['pre']>[1];
```
joshbwlng, line 30: This should just be `ActionRequestData`? These two types are actually incompatible at the moment, but we should use a single type for both action request contracts and the function we use to pre-process action requests.

joshbwlng:
@balena-ci retest

joshbwlng:
@balena-ci I self certify!

PR #8517, opened 2022-04-28
UI: Improve loop visualization

LucianBuzzo:
This change removes "projects", and changes the PR query to actually
return values. It also improves the labelling and includes a prototype
for calculating a rough average lifetime of improvements.

***

test suite!


LucianBuzzo:
@balena-ci I self-certify!

PR #8521, opened 2022-04-29
Remove Livepush library scope assumption

joshbwlng:
Until now our Livepush library setup has
assumed that all library package names
would be scoped under balena. This is
no longer the case for autumndb, so
removing this assumption.

Change-type: patch


joshbwlng:
@balena-ci I self certify!

joshbwlng:
@balena-ci I self certify!

PR #8543, opened 2022-05-01
Update unhandled server log call

joshbwlng:
logger.exception() requires the third argument
to be an Error object, but sometimes that is
not the case. A separate error about this is
then output, causing confusing logs.

Change-type: patch

***

test suite!


joshbwlng:
@balena-ci I self certify!

PR #8577, opened 2022-05-06
UI: Include dedicated improvements in count on user table

LucianBuzzo:
Previously the user table would only show the number of owned
improvements, but wouldn't include "dedicated" users, which indicates
they are working on an improvement but not the owner.

Change-type: minor

***

test suite!


LucianBuzzo:
@balena-ci I self-certify!

PR #8628, opened 2022-05-16
Remove stale contracts

ab77:
* kill stale? VersionBot contracts

Change-type: minor


ab77:
@balena-ci retest

ab77:
@joshbwlng @LucianBuzzo about those contracts... good to drop?

joshbwlng:
@ab77 From what I can tell, we should drop these. They all seem to reference very old (<=v28) versions of Jellyfish so obviously not in use at the moment. It looks like you're bringing in `.balena/secrets~HEAD` again with this PR though, which is unnecessary. Drop that from the PR and I'll approve.

joshbwlng:
@ab77 Dug into the PR changes a bit more and found these files have changed as well:
```
apps/grafana/provisioning/dashboards/logs.json
apps/grafana/provisioning/dashboards/metrics.json
apps/grafana/provisioning/datasources/datasources.yml
apps/prometheus/prometheus.yaml
balena.yml
secrets
```

I'm assuming the grafana/prometheus changes are bits that are no longer needed after the move to balena. Let me know if I'm wrong. The `secrets` change though is the changing the git submodule commit, so probably need to remove that from this PR.

Also found this for `balena.yml`:
```
diff --git a/balena.yml b/balena.yml
new file mode 100644
index 000000000..51bf3e77a
--- /dev/null
+++ b/balena.yml
@@ -0,0 +1,3 @@
+name: Jellyfish
+type: sw.application
+version: 50.2.4
```

Do you know if the version is automatically updated by versionbot?

ab77:
@joshbwlng the changes under Grafana is just removing the duplicates (but none of it is tested). Just cleaning up the obvious bits for now.

I am bumping the secrets (v3.0.42) since my personal key was added some time ago and I've now updated to use my company/balena one.

```
commit da1e079e86efa2979d4f555baf99d32cc6518d31 (origin/ab77/jellyfish-on-balena)
Author: ab77 <anton@balena.io>
Date:   Mon May 16 08:38:13 2022 -0700

    Remove personal key
    
    Change-type: patch
```

`balena.yml` version will (should) be handled by VersionBot.

joshbwlng (approved):

PR #8635, opened 2022-05-27
Add Reactions/Emojis

inwenc:
Change-type: minor

***

test suite!

image
image
image

### DIFF
--- a/apps/ui/lib/components/Event/Message/DisplayReactions.tsx
+++ b/apps/ui/lib/components/Event/Message/DisplayReactions.tsx
@@ -1,0 +1,135 @@
+import React, { useMemo } from 'react';
+import emoji from 'node-emoji';
+import styled from 'styled-components';
+import { useSelector } from 'react-redux';
+import { Button, Txt, useRequest } from 'rendition';
+import { useSetup } from '../../SetupProvider';
+import { selectors } from '../../../store';
+import { Contract } from 'autumndb';
+
+const Emoji = ({ symbol, count, ...rest }) => {
+	const em = React.useMemo(() => {
+		return emoji.emojify(symbol);
+	}, [symbol]);
+
+	return (
+		<Txt {...rest}>
+			{em}
+			<span>{count}</span>
+		</Txt>
+	);
+};
+
+const Wrapper = styled.section`
+	display: flex;
+	flex-direction: row;
+	align-content: space-between;
+	width: 200px;
+`;
+
+export const DisplayReactions = ({ message }: { message: Contract }) => {
+	const currentUser = useSelector(selectors.getCurrentUser());
+	const { sdk } = useSetup()!;
+
+	const reactions = useMemo(
+		() =>
+			message.links?.['has attached element']?.filter((contract) => {
+				return contract.type === 'reaction@1.0.0';
+			}),
+		[message],
+	);
+
+	const [groupReactions, , , forcePoll] = useRequest(
+		async () => {
+			if (!reactions) {
+				return {} as { [key: string]: Contract[] };
+			}
+			const reactionsArray = await sdk.query({
+				type: 'object',
+				$$links: {
+					'was created by': {
+						type: 'object',
+					},
+				},
+				properties: {
+					id: {
+						enum: reactions.map((reactionContract) => reactionContract.id),
+					},
+				},
+			});
+			return reactionsArray.reduce((map, reactionContract) => {
+				const symbol = reactionContract.data.reaction as string;
+				map[symbol] = map[symbol] || [];
+				map[symbol].push(reactionContract);
+				return map;
+			}, {} as { [key: string]: Contract[] });
+		},
+		[reactions],
+		{ polling: true },
+	);
+
+	// TODO: Remove this workaround when this is resolved
+	// https://github.com/product-os/autumndb/issues/1333
+
+	const handleClick = React.useCallback(
+		async (symbol) => {
+			const reactionByCurrentUser = currentUser
+				? (groupReactions ?? {})[symbol].find(
+						(reaction) =>
+							reaction.links?.['was created by']?.[0].id === currentUser['id'],
+				  )
+				: null;
+			if ((groupReactions ?? {})[symbol] && reactionByCurrentUser) {
+				await sdk.card.remove(
+					reactionByCurrentUser.id,
+					reactionByCurrentUser.type,
+				);
+			} else {
+				const newReaction = await sdk.card.create({
+					type: 'reaction',
+					name: symbol,
+					data: {
+						reaction: symbol,
+					},
+				});
+
+				await sdk.card.link(message, newReaction, 'has attached element');
+			}
+			forcePoll();
+		},
+		[groupReactions, message],
+	);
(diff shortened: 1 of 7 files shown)


inwenc:
@karaxuna - is this how you link to reactions?

apps/ui/lib/components/Event/Message/DisplayReactions.tsx
```diff
@@ -0,0 +1,165 @@
...
+					map[symbol].push(reaction.links?.['was created by']?.[0]!);
+					return map;
+				}, {} as { [key: string]: Contract[] });
+				setGroupReactions(result);
+			});
+	}, [reactions]);
+
+
+
+	// TODO: Enable this when this is resolved
```
LucianBuzzo, line 75: Can you please remove this commented-out code? Once the Autumndb issue is resolved we can review the code with fresh eyes. We're not really getting any benefit from leaving commented out code in this file.

apps/ui/lib/components/Event/Message/DisplayReactions.tsx
```diff
@@ -0,0 +1,165 @@
...
+
+	return (
+		<Wrapper>
+			{Object.entries(groupReactions).map(([symbol, users]) => {
+				return (
+					<Button onClick={() => handleClick(symbol)} p={1} mx={0.5}>
+						<Emoji
+							symbol={symbol}
+							count={users.length}
+							tooltip={{
```
LucianBuzzo, line 155: The tooltip should be placed on the `Button` element, so that you get the tooltip on hover of the control element, rather than just the inner `Emoji` component

LucianBuzzo (changes requested):
Nice work @inwenc - I've left a number of comments, but my main piece of feedback is that the current implementation is too slow. When you're adding or removing a reaction, having to wait for a round-trip to the API is adding too much latency. Can you find a way to avoid having to make the initial request to the API on click?
In theory you should already know who the creator of a reaction is before handling the `onClick` event for both adding and removing a reaction.

apps/ui/lib/components/Event/Message/Reactions.tsx
```diff
@@ -0,0 +1,97 @@
...
+		background: #eeeeee !important;
+	}
+`;
+
+export const Reactions = ({ message }) => {
+	const currentUser = useSelector(selectors.getCurrentUser());
+	const { sdk } = useSetup()!;
+	const onClick = React.useCallback(
+		async (symbol) => {
+			const [reaction] = await sdk.query({
```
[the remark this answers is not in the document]
inwenc: I just used one single query and then I broke it, @karaxuna fix it again. Then the query was done from the parent component and passed it as props to the displayReactions component and Reactions component. It became less stable and it's not much faster really. My previous version was more stable

apps/ui/lib/components/Event/Message/DisplayReactions.tsx
```diff
@@ -0,0 +1,165 @@
...
+	// 	console.log(map, reaction);
+	// 	const symbol = reaction.data.reaction as string;
+	// 	map[symbol] = map[symbol] || [];
+	// 	map[symbol].push(reaction.links?.['was created by']?.[0]!);
+	// 	return map;
+	// }, {} as { [key: string]: Contract[] });
+
+	const handleClick = React.useCallback(
+		async (symbol) => {
+			const [reaction] = await sdk.query({
```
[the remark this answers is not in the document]
inwenc: removed this query to use the query in the useEffect

inwenc:
@LucianBuzzo - could this PR merge into master?

LucianBuzzo:
@inwenc I'll review it first thing tomorrow morning 👍

apps/ui/lib/components/Event/Message/DisplayReactions.tsx
```diff
@@ -0,0 +1,134 @@
...
+
+	const { sdk } = useSetup()!;
+	// TODO: Remove this workaround when this is resolved
+	// https://github.com/product-os/autumndb/issues/1333
+	React.useEffect(() => {
+		if (!reactions) {
+			return;
+		}
+		const ac = new AbortController();
+		sdk
```
LucianBuzzo, line 51: Because `reactions` isn't memoized (e.g. with `React.memo`) this effect will loop infinitely. Everytime the group reactions are set in state, reactions will change and trigger this again.

LucianBuzzo: The reason for this is that when you use `.filter` on the message links, a new array is created, and the comparator in `useEffect` uses strict equality. The array's will no longer be strictly equal because `.filter` has created a new array and will re-run the callback.

apps/ui/lib/components/Event/Message/DisplayReactions.tsx
```diff
@@ -0,0 +1,134 @@
...
+			.query({
+				type: 'object',
+				$$links: {
+					'was created by': {
+						type: 'object',
+					},
+				},
+				properties: {
+					id: {
+						enum: reactions.map((reactionContract) => reactionContract.id),
```
LucianBuzzo, line 61: It's possible for `reactions` to be an empty array. You should check for this and exit early if there are no reactions

apps/ui/lib/components/Event/Message/DisplayReactions.tsx
```diff
@@ -0,0 +1,134 @@
...
+					type: 'reaction',
+					data: {
+						reaction: symbol,
+					},
+				});
+
+				await sdk.card.link(message, newReaction, 'has attached element');
+			}
+		},
+		[reactions, message],
```
LucianBuzzo, line 100: This callback does not have a dependency on `reactions` - you need to change this to `groupReactions`.
```suggestion
		[groupReactions, message],
```

Without this change, if you fix the infinite loop I commented on previously, then the onClick call back will not do anything, as `groupReactions` will always be the initial empty object.

apps/ui/lib/components/Event/Message/DisplayReactions.tsx
```diff
@@ -0,0 +1,134 @@
...
+	}, [reactions]);
+
+	const handleClick = React.useCallback(
+		async (symbol) => {
+			const reactionByCurrentUser = groupReactions[symbol].find(
+				(reaction) =>
+					reaction.links?.['was created by']?.[0].id === currentUser['id'],
+			);
+			if (groupReactions[symbol] && reactionByCurrentUser) {
+				await sdk.card.remove(
```
LucianBuzzo, line 85: Removing a reaction isn't working when I use a user with the community role, but I think this is an issue with the API permissions rather than a problem with your PR

apps/ui/lib/components/Event/Message/DisplayReactions.tsx
inwenc: In line 48, we are checking if reactions is empty and return on that condition

LucianBuzzo:
@inwenc I just tried this again and when I add a reaction the UI doesn't update

inwenc:
@LucianBuzzo - I just pushed up some changes.

LucianBuzzo (approved):

PR #8645, opened 2022-05-28
Update sinon from 11.1.2 to 14.0.0

joshbwlng:
Change-type: patch


joshbwlng:
@balena-ci I self certify!

PR #8648, opened 2022-05-30
Remove unused dev dependencies

karaxuna:
* Includes little fix to remove bluebird

Change-type: patch


karaxuna:
@balena-ci I self certify!

PR #8649, opened 2022-05-31
UI: Fix generation of kanban columns when viewing related contracts

LucianBuzzo:
Previously column generation would not work correctly when viewing
a collection of related contracts (as opposed to viewing a collection in
a view).

Notably, this fixes milestones and issues being displayed in kanban columns:

image


LucianBuzzo:
@balena-ci I self-certify!

PR #8650, opened 2022-05-31
Perform oauth with info from contracts

karaxuna:
When performing oauth flow against a provider (eg. login with balena), we used to get information from integration code. This information should be stored inside oauth-provider-contracts.

Change-type: major


karaxuna:
@balena-ci retest

joshbwlng (approved):

PR #8666, opened 2022-06-03
Server: Compress responses from http endpoints

LucianBuzzo:
This change adds the compression middleware to express, compressing
responses from the http API. This dramatically reduces the size of data
blobs downloaded. This is the same strategy and module used by
balenaCloud.

Change-type: patch

***

test suite!


LucianBuzzo:
@balena-ci I self-certify!

PR #8678, opened 2022-06-06
UI: Add support for dark mode

LucianBuzzo:
Change-type: patch

***

test suite!


LucianBuzzo:
image

LucianBuzzo:
This is an exploration into how to implement dark mode in Jellyfish. To get to the finish line with this, we need to remove all the existing places in the source code that hardcode colors and instead derive the colors from the top level theme.
We then need to switch theme based on the system settings.

PR #8680, opened 2022-06-06
Set timeouts to sane values

ab77:
* set tunnel timeout value (WSS)
* set http-keepalive timeout
* reflect AWS/ALB settings (defaults)

Change-type: patch


ab77:
@balena-ci I self-certify!

PR #8684, opened 2022-06-06
Revert recent changes to paid support thread view

LucianBuzzo:
The recent changes were more performant but surfaced up a lot of stale
zombie threads in production from 2018/2019. Closing these caused very
heavy load on the API servers, due to lots of sync activity.
This needs to be resolved separately, and then we can revisit query
optimisations for the view.

Change-type: patch

***

test suite!


LucianBuzzo:
@balena-ci I self-certify!

PR #8696, opened 2022-06-07
Fix iteration of support threads, now correctly handles discussions

ramirogm:
Change-type: patch

***

test suite!


ramirogm:
Fixes #8694

PR #8724, opened 2022-06-10
Fix getting types from view schema

karaxuna:
User can't fetch `ui-theme` type, so he gets `[ undefined ]` in `tailTypes` array that causes ui to crash.
This pr also removes duplicate code for getting types from schema.

Change-type: patch
See: https://www.flowdock.com/app/rulemotion/p-cyclops/threads/hIbAiZLV-aTP9pXiKY9Uq3416pD


karaxuna:
@balena-ci I self certify!

karaxuna:
@balena-ci rebase

PR #8726, opened 2022-06-10
Server: Improve heuristics when evaluating triggers

LucianBuzzo:
This brings in improvements to the worker module that make it better at
rejecting non-matching triggers without needing to make a DB query,
increasing performance.

Change-type: patch

***

test suite!


LucianBuzzo:
@balena-ci I self-certify!

PR #8727, opened 2022-06-10
Server: improve stream performance

LucianBuzzo:
Update to the latest autumnDB version which includes improvements to
stream performance, resulting in less queries to the DB when streaming
a query.

Change-type: patch

***

test suite!


LucianBuzzo:
@balena-ci I self-certify!

ghost (approved):

PR #8731, opened 2022-06-11
CI: Reduce test container build time

joshbwlng:
Change-type: patch

Reduce test container build time by baking required packages
into the base image instead of installing them on every build.


joshbwlng:
@balena-ci I self certify!

PR #8732, opened 2022-06-11
CI: Drop shellcheck install from test base image

joshbwlng:
Change-type: patch

We are installing shellcheck through npm and therefore
do not need to have it installed in the test base image
as well.


joshbwlng:
@balena-ci I self certify!

PR #8736, opened 2022-06-13
Server: Update to latest jellyfish-worker@30.0.9

LucianBuzzo:
This fixes an issue where some triggers would not run if they matched on
an inverse relationship name.

Change-type: patch

***

test suite!


LucianBuzzo:
@balena-ci rebase

PR #8745, opened 2022-06-14
Stream thread highlights

karaxuna:
We previously used thread's  `linked_at` property to detect updates and fetch highlights. Now we stream highlights directly.

Change-type: patch


karaxuna:
@balena-ci I self certify!

ghost (approved):

PR #8749, opened 2022-06-15
Use session identity instead of tokens for worker and kernel

LucianBuzzo:
This change brings in the [update to autumndb](https://github.com/product-os/autumndb/pull/1363) that removes direct session handling.
Instead of AutumnDB handling session tokens, it only deals with the identity/actor making requests. Authorization of users is now hoisted up to the route layer at the top of the call stack, where it can eventually be delegated to an external service, such as Ory Kratos.
This has the effect of only retrieving the actor once at the top of the call tree, rather than in every interaction with the autumnDB kernel interface, which brings a fairly hefty performance boost.

Change-type: major

***

test suite!


LucianBuzzo:
Running some simple benchmarking locally, this change brings a ~20% performance increase when creating a thread.

I used the following bench script:
```
const { getSdk } = require('@balena/jellyfish-client-sdk');
const bench = require('nanobench');

const setup = async () => {
	const sdk = getSdk({
		apiPrefix: 'api/v2',
		apiUrl: 'http://localhost:8000',
	});

	await sdk.auth.login({
		username: 'jellyfish',
		password: 'jellyfish',
	});
	return sdk;
};

bench('create thread 100 times', async function(b) {
	const times = 100;
	const sdk = await setup();

	b.start();

	for (var i = 0; i < times; i++) {
		// Create a thread with the SDK
		await sdk.card.create({
			type: 'thread@1.0.0',
		});
	}

	b.end();
});
```

On master this gives the following result:
```
ok ~26 s (25 s + 976374500 ns)
```

On this branch I get:
```
ok ~20 s (20 s + 148205800 ns)
```

LucianBuzzo:
@joshbwlng I would have used hyperfine here, but I had some trouble installing it due to missing deps on my machine

joshbwlng:
@LucianBuzzo We could make the switch to `nanobench` instead of that works better for everyone. I'd rather have one tool instead of multiple. And great work by the way, another ~20% performance boost is amazing! 👍

LucianBuzzo:
@joshbwlng I could probably get hyperfine working with a bit of effort, and it's language agnostic which is nice - maybe its a conversation for another time.
The performance gains are very nice - now I just need to get these e2e UI tests passing 😮‍💨

LucianBuzzo:
@karaxuna This PR is consistently failing on this livechat test:
```
[e2e-ui]   1) test/e2e/ui/livechat/index.spec.js:172:1 › Create conversation page ===========================
[e2e-ui] 
[e2e-ui]     Error: No notifications found
[e2e-ui] 
[e2e-ui]       213 |
[e2e-ui]       214 | 	// Should receive notification
[e2e-ui]     > 215 | 	const [ notification ] = await macros.waitForNotifications(page, 1)
[e2e-ui]           | 	                         ^
[e2e-ui]       216 | 	expect(notification.links['is attached to'][0].id).toEqual(response.id)
[e2e-ui]       217 |
[e2e-ui]       218 | 	// Should display support agent's username
[e2e-ui] 
[e2e-ui]         at /usr/src/jellyfish/test/e2e/ui/livechat/macros.js:190:10
[e2e-ui]         at Object.exports.retry (/usr/src/jellyfish/test/e2e/ui/livechat/macros.js:12:18)
[e2e-ui]         at /usr/src/jellyfish/test/e2e/ui/livechat/index.spec.js:215:27
[e2e-ui] 
```

Any ideas how I can debug this? Everything seems to work fine locally 🤔

karaxuna:
This notifications are set here: https://github.com/product-os/jellyfish/blob/master/apps/ui/lib/components/Livechat.tsx#L32

I can take a look this weekends.

LucianBuzzo:
@karaxuna It's been really hard to get an accurate local setup working. I got something going, but the `product` value wasn't set - when I made it optional everything worked as expected.

LucianBuzzo:
Ok, I found the issue. When the worker generates a new notification, it first creates session on behalf of the user that created the subscription. Unfortunately this session is being created using the session of the *triggering* user, not the privileged user, resulting in a session where the actor doesn't have any roles, because regular users can't see the roles of other users. 🤦

apps/ui/lib/components/ChatWidget/hooks/use-notification-watcher.ts
```diff
@@ -50,15 +50,19 @@ export const useNotificationWatcher = ({ onNotificationsChange }) => {
...
-								data: {
-									type: 'object',
-									required: ['product'],
-									properties: {
-										product: {
-											const: product,
-										},
-									},
-								},
+								data: product
```
ramirogm, line 53: Should this be the inverse ( negated )?
as in:
```
data: !product ? {
										type: 'object',
								 }
               : {
										type: 'object',
										required: ['product'],
										properties: {
											product: {
												const: product,
											},
										},
								 },
```

apps/server/lib/auth.ts
```diff
@@ -0,0 +1,193 @@
...
+
+	// If it's not captured by the checks above, the token is invalid
+	throw new Error('Invalid token');
+};
+
+export const authMiddleware =
+	(kernel: Kernel) => async (request, response, next) => {
+		const authorizationHeader = request.headers['authorization'];
+		if (!authorizationHeader) {
+			return response.status(401).json({
```
ramirogm, line 139: In this case, the server returns a 401 and when checking the actual values on a Basic auth a 400 is returned if the token is missing.
I think the 400 should be a 401 below.
And maybe we could return a more explicit message on the `data` specifying what was the error? ( no Auth header, for example )

ramirogm (commented):
looks much cleaner.
I left some comments and suggestions

LucianBuzzo:
@balena-ci I self-certify!

PR #8753, opened 2022-06-16
Stop using a guest user session

LucianBuzzo:
Guest user session is unnecessary, and creates a lot of cruft. This change drops the concept entirely. Anonymous users can use the /login endpoint to get a session token and then authenticate.


LucianBuzzo:
@balena-ci I self-certify!

PR #8755, opened 2022-06-16
Server: Drop Bluebird package

LucianBuzzo:
In all cases we've been able to simplify code by either using native
promises or switching to async/await, removing the need for the Bluebird
package.

Change-type: patch

***

test suite!


LucianBuzzo:
@balena-ci I self-certify!

ghost (approved):

PR #8772, opened 2022-06-18
Update jest packages

joshbwlng:
Change-type: patch

***

Still have some `apps/ui` unit tests to fix up, but I've been able to handle most of the issues.


joshbwlng:
@balena-ci I self certify!

PR #8773, opened 2022-06-19
Renovate: Ignore redis updates

joshbwlng:
Manually update redis image to keep the same version as production.

Change-type: patch

***

test suite!


joshbwlng:
@balena-ci I self certify!

PR #8781, opened 2022-06-21
Allow livechat subdomains

karaxuna:
Add wildcard subdomains to livechat. This will allow different users to have their own local storage instead of sharing.

Change-type: minor


karaxuna:
@balena-ci rebase

LucianBuzzo (approved):

PR #8790, opened 2022-06-22
Add whisper to milestone relationship

joshbwlng:
Bring in missing whisper to milestone relationship

Change-type: patch


joshbwlng:
@balena-ci I self certify!

PR #8792, opened 2022-06-23
Run UI service lint and unit tests in UI dockerfile

LucianBuzzo:
Currently these tests are run inside the SUT dockerfile, which means
that image needs to load all of the UI source code as well. With this
change, the UI unit and lint tests are run as part of the UI service build step.
This approach makes more sense and should reduce the size and build time
of the SUT container, which is currently the biggest bottleneck in the
stack's build step.

Change-type: patch

***

test suite!


package.json
```diff
@@ -31,7 +31,7 @@
     "docs:diagram": "nomnoml docs/diagrams/architecture.noml > docs/assets/architecture.svg",
     "docs:summary": "./scripts/architecture-summary.sh > ARCHITECTURE.md",
     "clean": "rimraf apps/*/packages/*",
-    "postinstall": "(cd apps/server && npm i); (cd apps/ui && npm i)",
```
joshbwlng, line 34: I find this quite useful during local development so I can get all `node_modules` set up with a single command.

LucianBuzzo: I'm sure, this could be replaced by `npm run install:all` or something similar.
We need a different strategy for the SUT container - right now it has to build the entire stack and is easily the slowest container to build (~5m). This PR starts to reduce some of the overhead.

LucianBuzzo: I've added an `install` script that does the same thing

LucianBuzzo:
@balena-ci rebase

LucianBuzzo:
@balena-ci rebase

LucianBuzzo:
@balena-ci rebase

LucianBuzzo:
@balena-ci rebase

LucianBuzzo:
@balena-ci I self-certify!

PR #8797, opened 2022-06-24
UI: Improve performance of component unit testing

LucianBuzzo:
This change turns off in-place typechecking when running unit test
files, dramatically improving test performance. Typechecking still
happens when the UI app is built, so we don't lose type security here,
and see a big improvement in performance.

See: https://github.com/kulshekhar/ts-jest/blob/main/website/docs/getting-started/options/isolatedModules.md

Change-type: patch

***

test suite!


LucianBuzzo:
@balena-ci I self-certify!

PR #8811, opened 2022-06-28
Drop @balena/jellyfish-plugin-channels

joshbwlng:
jellyfish-plugin-channels functionality has been merged into
jellyfish-worker, primarily because channels should be
first-class citizen contracts and not in a plugin

Change-type: patch


joshbwlng:
@balena-ci I self certify!

LucianBuzzo:
Nice!

PR #8812, opened 2022-06-28
Dev: Remove old compose commands

joshbwlng:
Drop a couple npm run compose commands as they are no longer
usable after the move to balena

Change-type: patch


joshbwlng:
@balena-ci I self certify!

PR #8828, opened 2022-07-05
UI: Don't allow unused variables

LucianBuzzo:
This change enforces stricter type checking in the UI, including checking for unused variables.

Change-type: patch

***

test suite!


LucianBuzzo:
@balena-ci I self-certify!

PR #8833, opened 2022-07-07
UI: Update channel with API data once loaded by ChannelRenderer

LucianBuzzo:
This change will update the channel data in redux with the full target contract so that it can be used for things like the page title.

Change-type: patch

***

test suite!


20k-ultra:
@balena-ci rebase

20k-ultra:
@balena-ci rebase

PR #8849, opened 2022-07-11
UI: Implement new NavBar component

LucianBuzzo:
A number of changes that transition us to a new version of the Jellyfish UI that better supports structured communication:
- The user profile/avatar, search and loop selector will now appear in a persistent navigation element at the top of the page.
- The loop selector will now also navigate the user to the selected loop
  - If the user then navigates away from that loop then the loop filtering will be removed automatically. 
  - Once a loop is selected, the user can then select a product of the loop, which will also perform navigation.
- Loops now have a "Threads" tab that aggregates all threads attached to products attached to the loop.
  - A "Product" in this context is a repository that has been linked to a loop as a product.
- Repositories now have a thread tab, that allows you to work with threads in relation to the repo

image

Connects-to: https://github.com/product-os/jellyfish/issues/8788
Change-type: minor

***

test suite!


apps/ui/lib/components/LoopSelector/index.tsx
```diff
@@ -0,0 +1,197 @@
...
+								id: {
+									const: activeLoop.id,
+								},
+							},
+						},
+					},
+				})
+				.then((result) => {
+					setProducts(result);
+				})
```
karaxuna, line 92: Plain promises to avoid self executing async function? 😄

apps/ui/lib/components/LoopSelector/index.tsx
```diff
@@ -0,0 +1,197 @@
...
+
+	const onProductChange = ({ value }) => {
+		if (activeLoop) {
+			if (value) {
+				history.push(`/${activeLoop.slug}/${value.slug}`);
+			} else {
+				history.push(`/${activeLoop.slug}`);
+			}
+		}
+	};
```
karaxuna, line 117:
```suggestion
	const onProductChange = React.useCallback(({ value }) => {
		if (activeLoop) {
			if (value) {
				history.push(`/${activeLoop.slug}/${value.slug}`);
			} else {
				history.push(`/${activeLoop.slug}`);
			}
		}
	}, [ activeLoop ]);
```

apps/ui/lib/components/LoopSelector/index.tsx
LucianBuzzo: Yep pretty much! I'm not sure about the pros/cons of using this instead of an IIFE and async/await - what are your thoughts?

apps/ui/lib/components/LoopSelector/index.tsx
LucianBuzzo: Done 👍

apps/ui/lib/components/LoopSelector/index.tsx
karaxuna: I don't think there's much practical difference, I just avoid plain promises to stick with async/await coding style.

karaxuna (approved):

PR #8853, opened 2022-07-12
Use wildcard message relationship

joshbwlng:
Change-type: patch

***

Replacing multiple `message-is-attached-to` relationships with a single wildcard `message-is-attached-to-any` relationship.


joshbwlng:
@balena-ci I self certify!

PR #8856, opened 2022-07-13
Server: Add support for COUNT_BY function in formulas

LucianBuzzo:
The `COUNT_BY` function is an alias for the lodash `_.countBy` function.

See https://lodash.com/docs/4.17.15#countBy

Change-type: patch

***

test suite!


LucianBuzzo:
@balena-ci I self-certify!

PR #8862, opened 2022-07-13
Ramirogm/async triggers 3

ramirogm:
***

test suite!


ramirogm:
@joshbwlng landrbot fails here, but this beta can be used to testing. I've tested this locally deployed and works as expected

joshbwlng:
@ramirogm Yeah, you can ignore Landrbot failures

PR #8871, opened 2022-07-16
UI: Fix link verbs in link modal

joshbwlng:
Remove odd hidden box wrapping link verb select component.
Remove possibility of duplicate link verbs from list.
Only add link verbs to the select list that are valid for the
selected target contract type.

Change-type: patch

---

**Before**
before-1
before-2

**After**
after-1
after-2

### DIFF
--- a/apps/ui/lib/components/LinkModal/LinkModal.tsx
+++ b/apps/ui/lib/components/LinkModal/LinkModal.tsx
@@ -2,20 +2,17 @@ import React from 'react';
 import _ from 'lodash';
 import { strict as assert } from 'assert';
 import pluralize from 'pluralize';
-import { Badge, Modal, Select, Txt, Box } from 'rendition';
+import { Badge, Modal, Select, Txt } from 'rendition';
 import * as notifications from '../../services/notifications';
 import { Icon } from '../';
 import type { Contract, ContractSummary, TypeContract } from 'autumndb';
 import { AutoCompleteCardSelect } from '../AutoCompleteCardSelect';
-import { Hideable } from '../Hideable';
 import * as linkUtils from './util';
 import { TypeFilter } from './TypeFilter';
 import * as helpers from '../../services/helpers';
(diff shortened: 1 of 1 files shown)


apps/ui/lib/components/LinkModal/LinkModal.tsx
```diff
@@ -190,20 +185,18 @@ export const LinkModal: React.FunctionComponent<Props> = ({
 				{cards[0].name || cards[0].slug}
 			</Txt>
 			{!linkVerb && (
-				<HideableBox isHidden={!selectedTarget || linkTypeTargets.length === 1}>
```
joshbwlng, line 193: @LucianBuzzo @karaxuna Not sure what the point of this `HideableBox` is, but please correct me if we need to keep it. The link modal works as I expect it to after removing it and only using the wrapped `Select` component.

LucianBuzzo: @joshbwlng IIRC the hideable box exists so that you can omit the `Select` if there is only 1 possible option for the relationship.

joshbwlng: @LucianBuzzo @karaxuna I was confused because while it does *kind of* hide the box, its actually still there if you click it:
HideableBox

joshbwlng: And I actually think it feels fine to have a select with only one option - a consistent feel as the options are filtered down when a target type and contract is selected.

joshbwlng: If everyone agrees I think we can merge this. Let me know if you think there's a better option though.

LucianBuzzo: Ok sure, lets merge it

LucianBuzzo:
@balena-ci rebase

LucianBuzzo (approved):

PR #8880, opened 2022-07-19
Drop plugin-default

joshbwlng:
Remove jellyfish-plugin-default as a dependency as it has been
merged into jellyfish-worker.

Change-type: patch


joshbwlng:
@balena-ci I self certify!

PR #8885, opened 2022-07-19
Server: Lint code as part of docker build

LucianBuzzo:
This change is part of continuing work to have the SUT container only
deal with e2e testing of the application instead of loading the source
code for the entire application so it can run lint and unit tests.

Change-type: patch

***

test suite!


LucianBuzzo:
@balena-ci I self-certify!

PR #8889, opened 2022-07-20
Remove duplicated type contracs that existed as test fixtures

Ereski:
Change-type: patch


Ereski:
@balena-ci I self-certify!

PR #8893, opened 2022-07-21
Server: Allow external support users to read relationships

LucianBuzzo:
This fixes a bug where external support users are unable to directly
create links between contracts because the SDK needs to read
relationships to create the correct link.

Change-type: patch

***

test suite!


joshbwlng (approved):

joshbwlng:
@balena-ci retest

PR #8901, opened 2022-07-22
Temporarily remove plugin-incidents

joshbwlng:
Removing plugin-incidents as adding it in caused heavy locking
in the production database on deployment. Will be adding this
plugin back in once the underlying cause has been fixed.

Change-type: patch

***

test suite!


joshbwlng:
@balena-ci I self certify!

PR #8924, opened 2022-07-28
Flowdock archive

iamsolankiamit:
Change-type: Major
Signed-off-by: Amit Solanki <amit@balena.io>

***

test suite!

### DIFF
--- a/apps/server/lib/bootstrap.ts
+++ b/apps/server/lib/bootstrap.ts
@@ -223,6 +223,147 @@ export const bootstrap = async (logContext: LogContext, options: any) => {
 		sync: worker.sync,
 	});
 
+	await worker.replaceCard(
+		logContext,
+		kernel.adminSession()!,
+		worker.typeContracts['type@1.0.0'],
+		{
+			attachEvents: false,
+			timestamp: new Date().toISOString(),
+		},
+		{
+			slug: 'flowdock-message',
+			type: 'type@1.0.0',
+			name: 'Flowdock message',
+			markers: [],
+			data: {
+				schema: {
+					required: ['data'],
+					type: 'object',
+					properties: {
+						data: {
+							required: [
+								'username',
+								'userFullName',
+								'flowdockUserId',
+								'flowdockUUID',
+								'sent',
+							],
+							type: 'object',
+							properties: {
+								flowdockUUID: {
+									type: 'string',
+								},
+								flowdockUserId: {
+									type: 'string',
+								},
+								userFullName: {
+									type: 'string',
+								},
+								username: {
+									type: 'string',
+								},
+								content: {
+									type: 'string',
+									format: 'markdown',
+									fullTextSearch: true,
+								},
+								sent: {
+									type: 'string',
+									format: 'date-time',
+								},
+								file: {
+									type: 'object',
+									properties: {
+										name: {
+											type: 'string',
+										},
+										mime: {
+											type: 'string',
+										},
+										bytesize: {
+											type: 'number',
+										},
+										slug: {
+											type: 'string',
+										},
+									},
+								},
+							},
+						},
+					},
+				},
+				indexed_fields: [['data.content']],
+			},
+		},
+	);
+
+	await worker.replaceCard(
+		logContext,
+		kernel.adminSession()!,
+		worker.typeContracts['type@1.0.0'],
+		{
+			attachEvents: false,
+			timestamp: new Date().toISOString(),
+		},
+		{
+			slug: 'flowdock-archive',
+			type: 'type@1.0.0',
+			name: 'Flowdock archive',
+			markers: [],
+			data: {
+				schema: {
+					required: ['data'],
+					type: 'object',
+					properties: {
+						data: {
+							required: ['title', 'flowdockThreadId', 'flowdockFlow'],
+							type: 'object',
+							properties: {
+								title: {
+									type: 'string',
+								},
+								flowdockThreadId: {
+									type: 'string',
+								},
+								flowdockFlow: {
+									type: 'string',
+								},
+							},
+						},
+					},
+				},
+				indexed_fields: [['data.flowdockThreadId']],
+			},
+		},
+	);
+
+	await worker.replaceCard(
+		logContext,
+		kernel.adminSession()!,
+		worker.typeContracts['relationship@1.0.0'],
+		{
+			attachEvents: false,
+			timestamp: new Date().toISOString(),
+		},
+		{
+			slug: 'relationship-flowdock-message-is-attached-to-flowdock-archive',
+			type: 'relationship@1.0.0',
+			name: 'is attached to',
+			data: {
+				inverseName: 'has attached element',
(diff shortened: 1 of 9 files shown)


apps/server/lib/bootstrap.ts
```diff
@@ -261,6 +261,64 @@ export const bootstrap = async (logContext: LogContext, options: any) => {
 		}),
 	);
 
+	await worker.insertCard(
```
iamsolankiamit, line 264: how do I  make this run only once? I have to keep on reseting the db, to make changes here.

LucianBuzzo: @iamsolankiamit You can try using `replaceCard`

LucianBuzzo:
@iamsolankiamit Whats the status on this? Is it ready for review?

iamsolankiamit:
Yes it is @LucianBuzzo apart from some styling issues everything is working as discussed

iamsolankiamit:
Screenshot 2022-08-19 at 12 35 16 PM
Screenshot 2022-08-19 at 12 35 27 PM

apps/server/lib/bootstrap.ts
```diff
@@ -275,6 +275,147 @@ export const bootstrap = async (logContext: LogContext, options: any) => {
...
+								content: {
+									type: 'string',
+									format: 'markdown',
+									fullTextSearch: true,
+								},
+								sent: {
+									type: 'string',
+									format: 'date-time',
+								},
+								file: {
```
iamsolankiamit, line 327: @LucianBuzzo I might be doing something wrong here, but files uploaded here are not being returned. Not sure what I am doing wrong.

apps/ui/lib/lens/misc/Flowdock/Messages.tsx
```diff
@@ -0,0 +1,181 @@
...
+	};
+};
+
+const INITIAL_ITEM_COUNT = 30;
+
+const START_INDEX = 10000;
+
+const DEFAULT_OPTIONS: SdkQueryOptions = {
+	limit: 30,
+	sortBy: 'created_at',
```
iamsolankiamit, line 85: I want to sort by 'data.sent' field in `flowdock-message` card. but the server is throwing errors. tried various things. couldn't pin point on the issue.

apps/ui/lib/lens/misc/Flowdock/Messages.tsx
```diff
@@ -0,0 +1,181 @@
...
+													@{message.data.username}({message.data.userFullName})
+												</Txt.span>
+											</Txt>
+										</Flex>
+									</HeaderWrapper>
+									<MessageContainer px={2} py={1} mb={1} mx={2}>
+										{message.data.content && (
+											<Markdown>{message.data.content}</Markdown>
+										)}
+										{/* {message.data.file && (
```
iamsolankiamit, line 162: the files i am uploading doesn't seems to be returned in the query. tried multiple types of queries but couldn't figureout the issue.

apps/server/lib/bootstrap.ts
iamsolankiamit: see here in the upload script https://github.com/balena-io-playground/flowdock-to-jf-uploader/blob/master/index.js#L90-L95

maybe i am doing something wrong there.

apps/server/lib/bootstrap.ts
```diff
@@ -275,6 +275,147 @@ export const bootstrap = async (logContext: LogContext, options: any) => {
 		}),
 	);
 
+	await worker.replaceCard(
```
LucianBuzzo, line 278: If these types are working as expected, can you add them to the `jellyfish-worker` module so they get loaded by default?

joshbwlng: @iamsolankiamit Seems like we should be able to use the generic `thread` and `message` contracts for this and avoid adding new types entirely. I really don't like the idea of adding custom cruft specifically for Flowdock if we can use types that already exist. If you need something to flag these items as "being from Flowdock", you should be able to add something under `data` for the contracts, `data.mirrors=["https://flowdock.com/flow-foo/thread-bar"]` for example.

iamsolankiamit: @joshbwlng I tried using the thread and message contract, but they need an actor and we need extra fields like `flowdockThreadId`,  `flowdockFlow`. I figured we could keep flowdock archive separate from other JF data.

iamsolankiamit: If you are more inclined towards using the same contracts. do you have suggestions on how to incorporate the additional data? @joshbwlng

joshbwlng: @iamsolankiamit What do you mean exactly when you say "an actor", that can have a few meanings in Jellyfish. I'm guessing you mean the user contract used to create the thread/message contracts? And you should be able to store anything extra you have under `data`, `data.flowdock.threadId` for example. Is there a reason for splitting up the parts of the flowdock URL into things like `flow` and `threadId`? Could these be parts of a single field? If not, whats the blocker?

If we find that we really need to have Flowdock-specific contract types, they shouldn't be defined here or in the worker, we would bring the `plugin-flowdock` repo back to life and modify it to do most if not all of what you've done in this PR (could probably copy-paste most over).

iamsolankiamit: @joshbwlng didn't know we could add extra data like like that into a contract. I won't be able to mark the fields as required though.

about splitting the `flow` and `threadId`. Flow names have changed in past so urls might contain older names. threadId is unique though so it won't change.

Yes by actor I mean the user contract. Let me try and use the existing Threads and messages contracts

joshbwlng: @iamsolankiamit I'm not sure we need to be able to mark these flowdock-only fields as required, we won't be making any new ones or updating them post-import correct? We use `required` to ensure contract data shape validity on upserts, but thats less important in this case from what I can tell.

And I'll trust your judgement on the `flow` and `threadId` fields, I think I understand what you're talking about. My only concern was unnecessarily duplicating data.

iamsolankiamit: @joshbwlng I am unable to create messages without an actor (i.e. the user) since some of the users are deleted. is there a way to skip the required property? Should we create a special user called deleted-user and fallback to that user when we don't know who posted that message?

joshbwlng: @iamsolankiamit I see the issue. What actor are you using now for creating the flowdock archive threads/messages in this PR? We might want to stick with more or less what you have now, but bring the implementation out of this repo and into a new plugin (or the defunct flowdock plugin I guess).

iamsolankiamit: @joshbwlng In the flowdock archive contract I made. I didn't make the actor field as required, instead I'd pass the flowdock username directly and name in the contract itself. 

While creating the contracts jf would assign the api caller as the actor. I think we can skip this by using the admin context

PR #8926, opened 2022-07-29
Update @balena/jellyfish-worker from 32.2.0 to 32.2.1

joshbwlng:
Add whisper to any relationship, as whispers can be included in
the timeline of any contract.

Change-type: patch


joshbwlng:
@balena-ci I self certify!

PR #8927, opened 2022-07-29
Update @balena/jellyfish-plugin-incidents from 7.2.1 to 7.2.2

joshbwlng:
Create ping messages on new incidents using hubot user if possible

Change-type: patch


joshbwlng:
@balena-ci I self certify!

PR #8941, opened 2022-07-31
Update ava from 4.3.0 to 4.3.1

joshbwlng:
Change-type: patch


joshbwlng:
@balena-ci I self certify!

PR #9016, opened 2022-08-05
UI: Refactor contracts to have multi-threaded conversation by default

LucianBuzzo:
By default contracts will have threads attached to them, that are
rendered in an interleaved style. The update and create events are now
displayed under and activity header in the info tab.
Sales threads, support threads and threads will be an exception as they
will need to render the message/whisper timeline.
Existing timeline messages will need to be migrated to fresh threads so
that they will still be discoverable.

Change-type: major

***

test suite!


LucianBuzzo:
@balena-ci rebase

LucianBuzzo:
@balena-ci I self-certify!

PR #9043, opened 2022-08-12
Clean up logger context

ramirogm:
Change-type: patch


ramirogm:
@balena-ci I self-certify!

PR #9073, opened 2022-08-15
UI: Default to providing the notification itself as the message context.

LucianBuzzo:
If a message thread is not attached to anything, use the notification
itself as  context, which allows the notification archive functionality
to work as expected.

Change-type: patch

***

test suite!


LucianBuzzo:
@balena-ci I self-certify!

PR #9088, opened 2022-08-16
UI: Makes 'Add [Support|Chat] Thread' floating to save screen real est...

aethernet:
Turned the Green add Thread button to floating and added some padding to the underlying scroll section to save some screen real estate and give an impression of space.


aethernet:
closed due to missing 
`Change-type: patch` in commit

PR #9094, opened 2022-08-17
UI: Makes 'Add [Support|Chat] Thread' floating

aethernet:
UI: Makes Add Thread button floating to save screen real estate.

Change-type: patch

### DIFF
--- a/apps/ui/lib/components/InfiniteList.tsx
+++ b/apps/ui/lib/components/InfiniteList.tsx
@@ -15,6 +15,7 @@ const ScrollArea = styled<any>(Box)`
 	overflow-y: auto;
 	display: flex;
 	flex-direction: column;
+	padding-bottom: 100px;
 `;
 
 const TRIGGEROFFSET = 200;
--- a/apps/ui/lib/lens/common/ViewFooter/ViewFooter.tsx
+++ b/apps/ui/lib/lens/common/ViewFooter/ViewFooter.tsx
@@ -14,7 +14,9 @@ import * as notifications from '../../../services/notifications';
 import { RelationshipContract } from 'autumndb';
 
 const Footer = styled(Flex)`
-	border-top: 1px solid #eee;
+	position: absolute;
+	bottom: 16px;
+	right: 16px;
 `;
 
 const DropUpButton = styled(DropDownButton).attrs({
--- a/apps/ui/lib/lens/common/ViewFooter/ViewFooter.tsx
+++ b/apps/ui/lib/lens/common/ViewFooter/ViewFooter.tsx
@@ -17,6 +17,7 @@ const Footer = styled(Flex)`
 	position: absolute;
 	bottom: 16px;
 	right: 16px;
+	z-index: 9;
 `;
 
 const DropUpButton = styled(DropDownButton).attrs({
@@ -139,7 +140,7 @@ export const ViewFooter: React.FunctionComponent<Props> = ({
 	);
 
 	return (
-		<Footer flex={0} p={3} {...rest} justifyContent="flex-end">
+		<Footer flex={0} mr={1} {...rest} justifyContent="flex-end">
 			{types.length > 1 ? (
 				<DropUpButton
 					alignRight
--- a/apps/ui/lib/lens/common/ViewFooter/ViewFooter.tsx
+++ b/apps/ui/lib/lens/common/ViewFooter/ViewFooter.tsx
@@ -166,13 +166,17 @@ export const ViewFooter: React.FunctionComponent<Props> = ({
 				<Button
 					disabled={isBusy}
 					success
+					width={38}
 					data-test={`viewfooter__add-btn--${types[0].slug}`}
 					onClick={() => {
 						handleAsyncError(onAddCard(types[0]));
 					}}
-				>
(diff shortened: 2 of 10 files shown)


joshbwlng:
@aethernet Could you include some before/after screenshots in your PR description to make it clear what "floating" means? Thanks!

apps/ui/lib/components/InfiniteList.tsx
```diff
@@ -15,7 +15,9 @@ const ScrollArea = styled<any>(Box)`
 	overflow-y: auto;
 	display: flex;
 	flex-direction: column;
+	padding-bottom: 100px;
```
LucianBuzzo, line 18: The additional bottom padding works fine here - its only an issue when you're at the most recent message

apps/ui/lib/lens/list/Interleaved/component.tsx
```diff
@@ -139,6 +139,20 @@ export const InterleavedList = (props: Props) => {
 			startReached={loadMoreContracts}
 			itemContent={itemContent}
 			overscan={10}
+			components={{
```
LucianBuzzo, line 142: What is this Footer component for?

apps/ui/lib/lens/common/ViewFooter/ViewFooter.tsx
```diff
@@ -14,7 +14,10 @@ import * as notifications from '../../../services/notifications';
 import { RelationshipContract } from 'autumndb';
 
 const Footer = styled(Flex)`
-	border-top: 1px solid #eee;
+	// border-top: 1px solid #eee;
```
LucianBuzzo, line 17: You can remove this commented out code

aethernet:
@joshbwlng here's
Cursor_and__52__3a563837-a2ad-4528-9559-5e4______Jellyfish
_52__3a563837-a2ad-4528-9559-5e4______Jellyfish
_52__6a3c84d9-7f5d-407f-a17b-917______Jellyfish

apps/ui/lib/lens/list/Interleaved/component.tsx
aethernet: Adds a last "empty" element on the virtuoso list, which have the same purpose as the margin-bottom.

We cannot easily attach the margin-bottom to any of the internal virtuoso divs and doing on the wrapper reduces the scrollview.

Next step for this footer is to add a spinner while loading and a "end of scroll" message (or image)

aethernet:
@joshbwlng before / after 

_52__6a3c84d9-7f5d-407f-a17b-917______Jellyfish-2

_52__3a563837-a2ad-4528-9559-5e4______Jellyfish-2

apps/ui/lib/lens/list/Simple/component.tsx
```diff
@@ -77,6 +77,20 @@ export default function SimpleList(props: Props) {
 			endReached={loadMoreContracts}
 			itemContent={itemContent}
 			overscan={5}
+			components={{
```
joshbwlng, line 80: Looks like you have some duplicate code

joshbwlng:
@aethernet Would these changes not result in a button that would block visibility of text/information under it?

aethernet:
@joshbwlng yes indeed, but it only matter if you cannot scroll to move the test out of under the button.

That's the reason for the margin-bottom (on generic scroll lists) / empty footer (on virtuoso). It ensure you can always scroll the last line out of the button (and gives a little bit of comfort at the end as the last line of the feed is not glue to the side of the window/screen)

aethernet:
_64__user-jmakivic___Jellyfish

apps/ui/lib/lens/common/ViewFooter/ViewFooter.tsx
```diff
@@ -14,7 +14,9 @@ import * as notifications from '../../../services/notifications';
 import { RelationshipContract } from 'autumndb';
 
 const Footer = styled(Flex)`
-	border-top: 1px solid #eee;
+	position: absolute;
```
LucianBuzzo, line 17: This causes problems for lenses that don't infinite scroll, like the table lens. We may need to conditionally fix the button based on the lens being rendered 🤔

LucianBuzzo (commented):
I've pushed a commit that cleans up the Footer a little. Overall I like the change, however the flooring button causes layout issues for more static lenses, like the table and we need to resolve that issue before this can be merged.

apps/ui/lib/lens/common/ViewFooter/ViewFooter.tsx
LucianBuzzo: After some though I think the best approach will be to give controls to lenses to declaratively not render the footer, and adjust the other lenses to compensate for a floating button.

LucianBuzzo:
@aethernet I tweaked this a bit more to make it play nicely with the other lenses, and added a similar affect for the segment footer (note to self: we need to unify those components). I also went a step further and switched to using an icon + tooltip for the buttons which makes them much smaller:

image

LucianBuzzo (approved):

aethernet:
@LucianBuzzo awesome !!

PR #9101, opened 2022-08-17
UI: Fix bug in custom view creation

LucianBuzzo:
This fixes an issue that would not save views as expected when in
a non-view context.

Change-type: patch

***

test suite!


LucianBuzzo:
@balena-ci I self-certify!

PR #9103, opened 2022-08-17
Rollback change to balena workflow that leaves dangling ec2 instances

ramirogm:
Change-type: patch


ramirogm:
@balena-ci I self-certify!

PR #9114, opened 2022-08-18
UI: Automatically mark message as read if it has an attached notification

LucianBuzzo:
Previously we would check the mentions* and alerts* fields, and only
mark a message as read if these fields contain the user slug. With the
switch to notifications we can simply check to see if the message has
attached notification for the current user.
This also means that marking a message as read when it is visible will
work for threads where the user is subscribed.

Change-type: patch

***

test suite!


LucianBuzzo:
@balena-ci rebase

LucianBuzzo:
@balena-ci rebase

LucianBuzzo:
@balena-ci I self-certify!

PR #9122, opened 2022-08-18
Strongly type Redux actions

LucianBuzzo:
This change introduces strong typing for redux actions, guaranteeing that actions use valid type names and values.


LucianBuzzo:
@balena-ci I self-certify!

PR #9127, opened 2022-08-19
Avoid restarting the jellyfish-tests /sbin/init

ramirogm:
Change-type: patch


ramirogm:
@balena-ci re-test

ramirogm:
@balena-ci re-test

ramirogm:
@balena-ci re-test

ramirogm:
Issue #9133

ramirogm:
@balena-ci re-test

ramirogm:
@balena-ci re-test

ramirogm:
@joshbwlng This is mainly a cleanup task. The details are explained on issue #9133

joshbwlng:
@ramirogm Thanks for the ping, I'll take a closer look when I'm back to work tomorrow. :+1:

.github/workflows/balena.yml
```diff
@@ -575,8 +575,7 @@ jobs:
 
               ssh-uuid -o StrictHostKeyChecking=no -o UserKnownHostsFile=/dev/null \
               ${{ steps.register-test-device.outputs.balena_device_uuid }}.balena \
-              'balena ps -q | xargs balena inspect | jq -r .[].State.Status' \
-                | grep -E 'created|restarting|removing|paused|exited|dead'
```
joshbwlng, line 579: What's the reason for this change? I'm assuming this changes the output seen in CI stdout used for debugging/logging purposes?

docker-compose.yml
```diff
@@ -23,10 +23,14 @@ services:
   jellyfish-tests:
     build:
       context: .
-      dockerfile: Dockerfile.template
+      dockerfile: Dockerfile.jf-tests
```
joshbwlng, line 26: I think you can leave the filename as `Dockerfile.template` as we're not using it anywhere else. `Dockerfile.template` also has a use/meaning in the balena world.

scripts/run-jf-tests.sh
```diff
@@ -0,0 +1,10 @@
+#!/bin/bash
+
+if [[ $SUT -eq 1 ]]
+then 
+  echo "SUT, running /usr/bin/entry.sh"
+  exec /usr/bin/entry.sh
+else
+  echo "No SUT, sleeping"
+  /bin/sleep infinity
+fi
```
joshbwlng, line 10: We already have a `scripts/ci/run-tests.sh`. The naming here will be confusing. I think this means we'd end up with a shell script (`scripts/run-jf-tests.sh`) that calls another shell script (`/usr/bin/entry.sh`) that then calls another shell script (`scripts/ci/run-tests.sh`)? A mess that is hard to follow and would like to clean up. What does `/usr/bin/entry.sh` do by the way? I wasn't aware of it before.

.github/workflows/balena.yml
ramirogm: Yes, it's a "verbose mode" to help debugging if the github workflow hangs there

docker-compose.yml
ramirogm: Ok, will revert.

scripts/run-jf-tests.sh
ramirogm: Oh it took me a while to untangle how this works! See #9133.
To keep it short, I want to use this script ( that can be renamed to "start-jellyfish-test-container ) to just idle the container in production.
The /usr/bin/entry.sh inherited from the Dockerfile at the openbalena-base-image does a `exec /sbin/init` that generates a lot of noise in the logs and is not really needed - the container is not used in prod

ramirogm: Also note that the `scripts/ci/run-tests.sh` is not called from the container, but instead is executed through ssh from the github workflow.

The script I wrote works as a wrapper for the inherited `/usr/bin/entry.sh`. Maybe it would be good to move both the `Dockerfile` and the script to their own folder to make it clear that they're only used to run tests as a kind of sidecar.

ramirogm: About >Dockerfile.template also has a use/meaning in the balena world.

In this case since the file is at the top level I think it's a bit confusing, at least for me, because it's not a `Dockerfile.template` as you would use on a mono-container balena app, but is something specific for running the tests. Wdyt of moving it to a subfolder `./apps/test-runner` ?

docker-compose.yml
ramirogm: Also check comment below

scripts/run-jf-tests.sh
joshbwlng: @ramirogm I agree with moving it out of the root directory, its confusing and acts as an odd special case for no good reason. Maybe we can rename the service in `docker-compose.yml` from `jellyfish-tests` to `test-runner` and then move the Dockerfile to `apps/test-runner` like you suggested. We'll have to keep the service's `build.context` as the repo root for now so it can access what it needs, but I think it might make sense to move everything test related to `apps/test-runner` in the future.

ramirogm:
@balena-ci re-test

ramirogm:
@balena-ci re-test

ramirogm:
@balena-ci re-test

ramirogm:
@balena-ci re-test

ramirogm:
@balena-ci I self-certify!

ramirogm:
Closes #9133

PR #9130, opened 2022-08-19
UI: add localTime on username tooltips

aethernet:
If a user have set his time zone, shows his localtime when hovering on it's name / avatar.
_69__8f61fd58-aa1d-4e1b-b161-4f6______Jellyfish


apps/ui/lib/services/helpers.ts
```diff
@@ -929,23 +929,39 @@ export const username = (slug: string) => {
...
 
+/** Given a timeZone string (i.e. `America/New-York` or `Europe/Brussels`) and an optional locale (defualt `en-US`), will return current time at that TZ */
+export const getLocalTimeForTimezone = (
+	timeZone: string,
+	locale?: string,
+): string | null => {
+	if (!timeZone) {
+		return null;
+	}
+	return new Date().toLocaleString(locale ?? 'en-US', { timeZone });
```
joshbwlng, line 940: @aethernet Looking good, but can we just set the first argument to `toLocaleString()` as `undefined`? From what I understand, when `undefined` the function will fall back to the users locale: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date/toLocaleString

Haven't tested with all browsers, but this is what I got in Chrome (`en-US` locale):
```
let now = new Date();

now.toLocaleString(undefined, { timeZone: 'Europe/London' })
'8/21/2022, 9:55:41 AM'

now.toLocaleString('en-US', { timeZone: 'Europe/London' })
'8/21/2022, 9:55:41 AM'

now.toLocaleString('ja-JP', { timeZone: 'Europe/London' })
'2022/8/21 9:55:41'
```

joshbwlng:
@aethernet What do you think about having some placeholder text, something like `(Time zone not set)`, when users don't have a time zone set on their account. I can see people being confused when they can see timestamps for some people but not others.

apps/ui/lib/services/helpers.ts
aethernet: I'll retry that, i though i did and it crashed on me, but maybe it did it differently

aethernet:
@joshbwlng good idea

apps/ui/lib/services/helpers.ts
aethernet: Works fine, don't know why it didn't the first time.

aethernet:
next up would be to link it to the team model, but it's out of my current abilities :D

aethernet:
added the city if know which imho made the time information more explicit.
thought of adding the country as well, but it seemed too much

aethernet:
and we need to give a location and TZ to hubot

joshbwlng:
@aethernet I wouldn't worry about linking to the team model etc. We can circle back around and add more information in the future, this is good enough that I think people will be pretty happy to see it. Can you update your screenshot and get your commits sorted so that versionbot is happy? And I'm not sure why we need to give a city/time zone to hubot? I guess we could pick something, but it shouldn't be necessary for these changes to work.

aethernet:
@joshbwlng 
Absolutely, i was using `next up` in a very long term way :)

About Hubot, it's just to give it a little bit of personality :)

Here's the new screenshot 
Cursor_and_milestone-de-risk-applicati______Jellyfish

aethernet:
@joshbwlng i don't understand what i'm doing wrong with versionist :(

joshbwlng:
@aethernet Try squashing all of your commits into the first, I'd prefer that be done anyway to keep our commit history as clean as possible

aethernet:
I'll create PR from a cleaner branch

aethernet:
new PR is here : https://github.com/product-os/jellyfish/pull/9186

PR #9162, opened 2022-08-23
UI: Fix layout issues for linkable contracts when creating a new contract

LucianBuzzo:
This fixes a bug where the buttons to add or link a new contract on creation would obscure the submit button at the bottom of the screen.

Change-type: patch

***

test suite!


LucianBuzzo:
@balena-ci I self-certify!

PR #9190, opened 2022-08-26
Fix inbox to handle sent tab query results

joshbwlng:
Update the EventBox component used in the new Inbox to properly
handle cases in which the query results in a message/whisper contract
and stop assuming result is always a thread. Also add prop to toggle
archive button - should only be able to archive things from the open
tab.

Change-type: patch


joshbwlng:
@balena-ci rebase

joshbwlng:
@balena-ci I self certify!

PR #9201, opened 2022-08-29
UI: Prefer linked target to hardcoded target when clicking a message

LucianBuzzo:
This makes the on-click behaviour much more responsive to context than
it used to be, by relying on the linked data if it's available.
For example, the message may have initially targeted and entity that no
longer exists, but is being queried through a more recent link. By using
the link data, it is much more likely that clicking the message will
result in the correct behaviour.

Change-type: patch

***

test suite!

### DIFF
--- a/apps/ui/lib/components/Event/Message/index.tsx
+++ b/apps/ui/lib/components/Event/Message/index.tsx
@@ -53,8 +53,10 @@ const MessageIconWrapper = styled(Box)`
 	}
 `;
 
-const getTargetId = (card: Contract<{ target: string }>) => {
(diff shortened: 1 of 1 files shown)


LucianBuzzo:
@balena-ci rebase

LucianBuzzo:
@balena-ci I self-certify!

LucianBuzzo:
@balena-ci rebase

LucianBuzzo:
@balena-ci rebase

LucianBuzzo:
@balena-ci rebase

PR #9202, opened 2022-08-29
Correctly parse comments with attachments and no text

ramirogm:
Bumped from jellyfish-plugin-front 6.0.67

Change-type: patch


ramirogm:
@balena-ci I self-certify!

PR #9208, opened 2022-08-30
Improve notification sorting in inbox

LucianBuzzo:
This change ensures that the most recent message in a thread is shown and that the threads are sorted by order of most recent message.
Practically this means that you should see the most recent messages first, and at the top of the inbox.

### DIFF
--- a/apps/ui/lib/lens/misc/Inbox/InboxTab.tsx
+++ b/apps/ui/lib/lens/misc/Inbox/InboxTab.tsx
@@ -29,7 +29,7 @@ const getArchivedNotificationQuery = (): JsonSchema => ({
 });
 
 const DEFAULT_OPTIONS: SdkQueryOptions = {
-	limit: 15,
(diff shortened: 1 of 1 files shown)


LucianBuzzo:
@balena-ci rebase

LucianBuzzo:
@balena-ci I self-certify!

aethernet:
@LucianBuzzo 
when you say "most recent message" you mean "most recent notification" in the thread or any message ?

I tried to fix this one with a query but all my attempts failed to provide all the functionalities (especially the `+ x messsages`) without some sort of hacks.

When you have some time, I'd like to understand what would be the right solution for this.

aethernet:
(and thanks for the fix! this is probably the current number one frustration among the team)

LucianBuzzo:
@balena-ci rebase

LucianBuzzo:
@balena-ci rebase

PR #9221, opened 2022-08-31
Change navbar color and contrast

inwenc:
Change-type: minor

image
image

test suite!


inwenc:
@balena-ci I self-certify!

PR #9222, opened 2022-08-31
Retry if a query times out to avoid inconsistencies between front and JF

ramirogm:
Bump jellyfish-plugin-front to v6.0.73

Change-type: patch


ramirogm:
@balena-ci I self-certify!

PR #9225, opened 2022-09-01
Add another row of emojis and rearrange order

inwenc:
Change-type: minor
Signed-off-by: YingwenChen yingwenchen@balena.io

***
image

test suite!


inwenc:
@balena-ci I self-certify!

PR #9234, opened 2022-09-07
Fix : "allGroupsWithUsersQuery" in bootstrap takes 60mb and 10+ sec

aethernet:
Removed "allGroupsWithUsersQuery" from bootstrap.
That query was ~60mb and took about >10sec to load.

The group functionality seems fine as the data are also loaded in `/jellyfish/apps/ui/lib/components/ChatWidget/store/action-creators.ts` and requested when required.

The proper request should be ~45kb.

change-type: minor

### DIFF
--- a/apps/ui/lib/store/actioncreators/index.ts
+++ b/apps/ui/lib/store/actioncreators/index.ts
@@ -54,33 +54,6 @@ export type JellyThunkDispatch = ThunkDispatch<State, ThunkExtraArgs, Action>;
 // Refresh the session token once every 3 hours
 const TOKEN_REFRESH_INTERVAL = 3 * 60 * 60 * 1000;
 
-const allGroupsWithUsersQuery: JsonSchema = {
-	type: 'object',
-	description: 'Get all groups with member user slugs',
-	required: ['type', 'name'],
-	$$links: {
-		'has group member': {
-			type: 'object',
-			required: ['slug', 'active'],
-			properties: {
-				active: {
-					type: 'boolean',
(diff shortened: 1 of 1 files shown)


apps/ui/lib/store/actioncreators/index.ts
```diff
@@ -550,7 +522,6 @@ export const actionCreators = {
 						dispatch(actionCreators.setTypes(types));
 						dispatch(actionCreators.setRelationships(relationships));
 						dispatch(actionCreators.setOrgs(orgs));
-						dispatch(actionCreators.setGroups(groups, user));
```
LucianBuzzo, line 553: Do we use `setGroups` anywhere else?

aethernet: We do in [/jellyfish/apps/ui/lib/components/ChatWidget/components/Layout.tsx](https://github.com/product-os/jellyfish/blob/ff0799440f08020422d62b12dce64796aebcca96/apps/ui/lib/components/ChatWidget/components/Layout.tsx#L66)

LucianBuzzo: I think that uses a seperate redux store in the Chat widget component

LucianBuzzo:
@aethernet There are a bunch of places that use retrieve the `groups` value from the redux store using the `getGroups` selector. At first glance, it seems they would be affected by this change. If it turns out that they aren't affected, then the entire `groups` data store and it's associated getters and setters should be surgically extracted.

PR #9239, opened 2022-09-07
UI: Improve performance of chat group query

LucianBuzzo:
Only return the fields required in the chat group query

Change-type: patch

***

test suite!

### DIFF
--- a/apps/ui/lib/store/actioncreators/index.ts
+++ b/apps/ui/lib/store/actioncreators/index.ts
@@ -57,11 +57,10 @@ const TOKEN_REFRESH_INTERVAL = 3 * 60 * 60 * 1000;
 const allGroupsWithUsersQuery: JsonSchema = {
 	type: 'object',
 	description: 'Get all groups with member user slugs',
-	required: ['type', 'name'],
 	$$links: {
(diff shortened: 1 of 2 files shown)


apps/ui/lib/store/actioncreators/index.ts
```diff
@@ -543,6 +548,9 @@ export const actionCreators = {
 				]).then(async ([loops, orgs, types, relationships, groups, config]) => {
 					const state = getState();
 
+					console.log({ allGroupsWithUsersQuery });
+					console.log({ groups });
+
```
aethernet, line 553: you forgot those

aethernet (commented):
you left a console.log
this reduces the payload to 167k (instead of 67m), well done.

apps/ui/lib/store/actioncreators/index.ts
LucianBuzzo: resolved

LucianBuzzo:
I've removed the logs now, I think we hit some intermittent test failure. Rebuilding now

LucianBuzzo:
@balena-ci I self-certify!

PR #9246, opened 2022-09-09
Move Tooltip on user avatar to the right 

aethernet:
The tooltip was opening to the top from the avatar which is top left of the screen, so always out of the screen.
Now opening to the right so the information are readable.


joshbwlng (approved):

joshbwlng:
image

joshbwlng:
@balena-ci rebase

PR #9258, opened 2022-09-13
UI: Truncate bookmark names

joshbwlng:
Change-type: patch

***

Ideally we should have a more elegant solution, but this is better than what we have now and allows users to bookmark contracts that have very long names without breaking their sidebars.


joshbwlng:
@balena-ci I self certify!

PR #9260, opened 2022-09-13
Revert "UI: Fix unlink modal query"

joshbwlng:
Change-type: patch

***

The "fix" did fix one thing, but broke another. Need to spend more time figuring out a proper solution and add tests to ensure things don't break again. Will do this over the next few days.


joshbwlng:
@balena-ci I self certify!

PR #9271, opened 2022-09-14
UI: Improve image rendering

LucianBuzzo:
This change keeps the generated blob source data in cache in the
browser, reducing server load and visually disruptive re-renders.
This cache is ephemeeral and flushed on a browser navigation.
The ideal solution is to use cookies to authenticate files instead of
fetching blob data and assembling it in the browser.

Change-type: patch

***

Caches images posted in messages, instead of reloading them from the API every time. This fixes distracting resize events in timelines and interleaved lenses.


LucianBuzzo:
@balena-ci rebase

LucianBuzzo:
@balena-ci rebase

joshbwlng:
Uploaded images are currently not being resized

joshbwlng:
Things look to be working as expected, but need to double-check code changes one more time just in case

joshbwlng:
@balena-ci rebase

joshbwlng (approved):

PR #9274, opened 2022-09-14
Improve logging; add NODE_ENV to be able to filter out e2e tests from logs

ramirogm:
Change-type: patch


ramirogm:
@balena-ci I self-certify!

PR #9281, opened 2022-09-15
UI: Remove compacting of notifications in inbox

LucianBuzzo:
Each message you are notified of will now be a seperate item in the inbox. This change has been made to avoid the situation where notifications are missed, because they are obscured by a subsequent notification on the same context.

Change-type: patch

***

test suite!


LucianBuzzo:
@balena-ci I self-certify!

PR #9304, opened 2022-09-17
CI: Merge test image into test-runner

joshbwlng:
The Jellyfish test image is only used here, so merging it in and deprecating the product-os/jellyfish-base-images repository.

Change-type: patch


joshbwlng:
@balena-ci I self certify!

PR #9313, opened 2022-09-18
Dev: Switch to official MinIO image for S3

joshbwlng:
The open-balena-s3 image uses systemd and causes issues with systems
running cgroups v2. The official MinIO images do not have this problem.

Change-type: patch


joshbwlng:
@balena-ci I self certify!

PR #9355, opened 2022-09-22
Display direct pings only in inbox

inwenc:
Change-type: minor

***
image

test suite!

### DIFF
--- a/apps/ui/lib/components/Event/index.tsx
+++ b/apps/ui/lib/components/Event/index.tsx
@@ -18,7 +18,7 @@ const EventWithActor: React.FunctionComponent<any> = (props) => {
 	const mentionsUserArray = card.data.payload?.mentionsUser || [];
 	const directMessageMarkersArray = card.markers || [];
 	const isDirectMessagePing =
-		directMessageMarkersArray[0]?.split('+')[0] === user.slug || false;
+		directMessageMarkersArray[0]?.split('+').includes(user.slug) || false;
 	const isDirect = mentionsUserArray.includes(user.slug) || isDirectMessagePing;
 	const typeBase = props.card.type.split('@')[0];
 	if (isDirectMention) {
--- a/apps/ui/lib/components/Event/index.tsx
+++ b/apps/ui/lib/components/Event/index.tsx
@@ -1,4 +1,4 @@
-import React, { useState } from 'react';
+import React from 'react';
 import { compose } from 'redux';
 import type { UserContract } from 'autumndb';
 import { parseMessage } from './Message/Body';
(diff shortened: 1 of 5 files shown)


apps/ui/lib/lens/misc/Inbox/Inbox.tsx
```diff
@@ -122,14 +122,27 @@ const Inbox = ({ channel }: Props) => {
...
+						isDirectPing={false}
+					/>
 				</Tab>
 				<Tab title="Archived">
 					<InboxTab query={archivedQuery} channel={channel} />
 				</Tab>
 				<Tab title="Sent">
 					<InboxTab query={sentQuery} channel={channel} />
 				</Tab>
+				<Tab title="Direct Pings">
```
joshbwlng, line 138: So we're going with a seperate tab? Not a problem, but did you check in with @JonJRich and/or @Jazzagi just in case?

inwenc: @joshbwlng They both said new tab is good

apps/ui/lib/components/Event/index.tsx
```diff
@@ -14,35 +14,78 @@ import Message from './Message';
 export { parseMessage };
 
 const EventWithActor: React.FunctionComponent<any> = (props) => {
-	const { card, user, onCardVisible, targetCard } = props;
+	const { card, user, onCardVisible, targetCard, isDirectMention } = props;
+	const mentionsUserArray = card.data.payload?.mentionsUser || [];
+	const directMessageMarkersArray = card.markers || [];
+	const isDirectMessagePing =
+		directMessageMarkersArray[0]?.split('+')[0] === user.slug || false;
```
joshbwlng, line 21: Are there cases in which `card.markers` does have a `+`? If so, then that means multiple entities are assigned in the `markers` and we probably can't assume its always the first in the list (like you're doing with `[0]`.

apps/ui/lib/components/Event/index.tsx
```diff
@@ -14,35 +14,78 @@ import Message from './Message';
...
-				if (typeBase === LINK) {
-					return (
-						<LinkedCard actor={actor} card={card} targetCard={targetCard} />
-					);
-				}
-				return <Message {...props} actor={actor} />;
-			}}
-		</CardLoader>
-	);
+	if (isDirectMention) {
```
joshbwlng, line 24: This is a bit confusing for me, you have `isDirectMention` and `isDirect` used in conditionals, but without comments its very hard to read what's going on and why.

joshbwlng: Also most of the contents in the `else` block is a copy of whats in the `if` block. This should be consolidated to reduce duplicate code.

apps/ui/lib/lens/misc/Inbox/Inbox.tsx
```diff
@@ -122,13 +122,34 @@ const Inbox = ({ channel }: Props) => {
...
+					<InboxTab
+						query={openQuery}
+						channel={channel}
+						canArchive={true}
+						isDirectMention={false}
+					/>
+				</Tab>
+				<Tab title="Direct Mentions">
+					<InboxTab
+						query={openQuery}
```
joshbwlng, line 134: This query will return a lot of things that are not direct mentions, when I believe you are then filtering on. Are you sure there's not a better query we can use to reduce work on both the backend and frontend (not to mention network payload size).

joshbwlng:
@inwenc Left a few comments. Also, can you try and add an e2e ui test (under `test/e2e/ui`) to show that direct mentions, and only direct mentions, are displayed in the new tab? You will also need to squash your commits.

apps/ui/lib/components/Event/index.tsx
inwenc: it's because there's 2 booleans to keep track, if isDirectMention then display the filtered cards from the isDirect check. Else render all @ group and @ user-slug

joshbwlng: I understand the general logic flow, what I'm asking is that you reduce duplicate code and at least add comments to explain what's happening.

joshbwlng: Having clearer variable names may help here as well, `onlyDirectMentions` boolean instead of `isDirectMention` for example.

joshbwlng:
@inwenc Added a few e2e-ui tests, looks like direct alerts (`!joshbwlng` instead of `@joshbwlng` for example) aren't showing up in the new tab

apps/ui/lib/components/Event/index.tsx
```diff
@@ -14,35 +14,82 @@ import Message from './Message';
 export { parseMessage };
 
 const EventWithActor: React.FunctionComponent<any> = (props) => {
-	const { card, user, onCardVisible, targetCard } = props;
+	const { card, user, onCardVisible, targetCard, isDirectMention } = props;
+	const mentionsUserArray = card.data.payload?.mentionsUser || [];
```
joshbwlng, line 18: If I remember correctly, you'll need to also check `alertsUser` for alerts

apps/ui/lib/lens/misc/Inbox/Inbox.tsx
joshbwlng: I have an idea on a query that might do the trick. I'll try it either this weekend or Monday and let you know how it goes. 👍

joshbwlng: Trying my idea here: https://github.com/product-os/jellyfish/pull/9357

joshbwlng:
Went with a different, simpler strategy here: https://github.com/product-os/jellyfish/pull/9357

PR #9360, opened 2022-09-26
Revert "Include private messages in direct mentions tab"

joshbwlng:
Change-type: patch

***

Reverts product-os/jellyfish#9359

Query causes timeouts right now because the database is too large (works fine in development/CI). Going with different strategy for the time being.


joshbwlng:
@balena-ci I self certify!

ghost (approved):

PR #9362, opened 2022-09-27
Add SQL to archive old contracts

joshbwlng:
Change-type: patch

***

Add SQL to create archive tables and then move old contracts to these archive tables. Should run this regularly with [pg_cron](https://docs.aws.amazon.com/AmazonRDS/latest/UserGuide/PostgreSQL_pg_cron.html#PostgreSQL_pg_cron.enable), once a month?

### DIFF
--- a/sql/archive-tables.sql
+++ b/sql/archive-tables.sql
@@ -1,0 +1,7 @@
+SET statement_timeout=0;
+
+-- Create archive_cards table
+CREATE TABLE IF NOT EXISTS public.archive_cards (LIKE cards INCLUDING ALL);
+
+-- Create archive_links2 table
+CREATE TABLE IF NOT EXISTS public.archive_links2 (LIKE links2 INCLUDING ALL);
--- a/sql/archive.sql
+++ b/sql/archive.sql
@@ -1,0 +1,48 @@
+SET statement_timeout=0;
+
+-- Move old contracts to archive tables
+DO $$
+DECLARE
+    contract_ids UUID[];
+    link_ids UUID[];
+    primary_link_ids: UUID[];
+    secondary_link_ids: UUID[];
+BEGIN
+    LOOP
+        -- Select contracts we no longer need in the main table
+        SELECT array_agg(id) INTO contract_ids FROM (
+            SELECT id FROM cards WHERE (type='execute@1.0.0' OR (type='session@1.0.0' AND slug!='session-admin-kernel') OR type='external-event@1.0.0' OR active=false)
+                AND created_at < to_char(CURRENT_DATE - INTERVAL '3 months', 'YYYY-MM-DD')::date LIMIT 1000
+        ) as subset;
+        IF (contract_ids IS NULL) THEN
+            RAISE INFO 'No more matching contracts found, quitting...';
+            EXIT;
+        END IF;
+        RAISE INFO 'Archiving % contracts...', array_length(contract_ids, 1);
+
+        -- Select links for matching contracts
+        SELECT array_agg(id) INTO primary_link_ids FROM links2 WHERE fromid = ANY(contract_ids) OR toid = ANY(contract_ids);
+        SELECT array_agg(id) INTO secondary_link_ids FROM links2 WHERE fromid = ANY(primary_link_ids) OR toid = ANY(primary_link_ids);
+        link_ids = array_cat(primary_link_ids, secondary_link_ids);
+        RAISE INFO 'Archiving % links...', array_length(link_ids, 1);
+
+        -- Archive link records from links2 table
+        WITH moved_rows AS (
+            DELETE FROM links2 WHERE id = ANY(link_ids) OR fromid = ANY(link_ids) OR toid = ANY(link_ids) RETURNING *
+        ) INSERT INTO archive_links2 SELECT * FROM moved_rows;
+
+        -- Archive link records from cards table
+        WITH moved_rows AS (
+            DELETE FROM cards WHERE type='link@1.0.0' AND id = ANY(link_ids) RETURNING *
+        ) INSERT INTO archive_cards SELECT * FROM moved_rows;
+
+        -- Archive contract records from cards table
+        WITH moved_rows AS (
+            DELETE FROM cards WHERE id = ANY(contract_ids) RETURNING *
+        ) INSERT INTO archive_cards SELECT * FROM moved_rows;
+
+        -- Sleep for 1 second so as to not overload the database
+        PERFORM pg_sleep(1);
+    END LOOP;
+END;
+$$ LANGUAGE plpgsql;
--- a/sql/generate.sql
+++ b/sql/generate.sql
@@ -1,0 +1,301 @@
+CREATE EXTENSION IF NOT EXISTS hstore;
+DO $$
+DECLARE
+    link_map HSTORE;
+    link_names TEXT[];
+BEGIN
+    link_map = hstore(
+        ARRAY['has attached element','is attached to','is creator of','was created by','executes','is executed by'],
+        ARRAY['is attached to','has attached element','was created by','is creator of','is executed by','executes']
+    );
+
+    -- Clear important bits
+    TRUNCATE links2, cards RESTART IDENTITY;
+
+    -- Disable triggers or this will go on for ages
+    SET session_replication_role = replica;
+
+    -- Seed 250k user contracts
+    RAISE INFO 'Creating user contracts...';
+    INSERT INTO cards (id, slug, type, active, version_major, version_minor, version_patch, tags, markers, links, requires, capabilities, data, linked_at, created_at)
+    SELECT
+        uuid_in(md5(random()::text || clock_timestamp()::text)::cstring),
+        'user-' || uuid_in(md5(random()::text || clock_timestamp()::text)::cstring),
+        'user@1.0.0',
+        random() > .01,
+        1,
+        0,
+        0,
+        '{}',
+        '{}',
+        '{}',
+        '{}',
+        '{}',
+        ('{"hash": "' || md5(random()::text || clock_timestamp()::text) || '", "email": "' || md5(random()::text || clock_timestamp()::text) || '@' || md5(random()::text || clock_timestamp()::text) || '.' || md5(random()::text || clock_timestamp()::text) || '"}')::json,
+        '{}',
+        (NOW() + (random() * (interval '90 days')) + '30 days')
+    FROM generate_series(1, 250000);
+
+    -- Create 1.5 session cards for each user
+    RAISE INFO 'Creating session contracts...';
+    INSERT INTO cards (id, slug, type, active, version_major, version_minor, version_patch, tags, markers, links, requires, capabilities, data, linked_at, created_at)
+    SELECT
+        uuid_in(md5(random()::text || clock_timestamp()::text)::cstring),
+        'session-' || uuid_in(md5(random()::text || clock_timestamp()::text)::cstring),
+        'session@1.0.0',
+        random() > .01,
+        1,
+        0,
+        0,
+        '{}',
+        '{}',
+        '{}',
+        '{}',
+        '{}',
+        ('{}')::json,
+        '{}',
+        (NOW() + (random() * (interval '90 days')) + '30 days')
+    FROM generate_series(1, (
+        SELECT floor(1.5 * count(*))::int
+        FROM cards
+        WHERE cards.type = 'user@1.0.0'
+    ));
+
+    -- Randomly link sessions to users
+    RAISE INFO 'Linking session contracts to user contracts...';
+    WITH users AS MATERIALIZED (
+        SELECT array_agg(cards.id) AS ids
+        FROM cards
+        WHERE cards.type = 'user@1.0.0'
+        ORDER BY random()
+    )
+    INSERT INTO links2 (id, forward, fromid, name, toid)
+    SELECT
+        uuid_in(md5(random()::text || clock_timestamp()::text)::cstring),
+        true,
+        users.ids[(random() * (array_length(users.ids, 1) - 1) + 1)::int],
+        (SELECT id FROM strings WHERE string='is creator of'),
+        cards.id
+    FROM cards, users
+    WHERE cards.type = 'session@1.0.0';
+
+    WITH users AS MATERIALIZED (
+        SELECT array_agg(cards.id) AS ids
+        FROM cards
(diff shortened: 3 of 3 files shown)


sql/archive.sql
```diff
@@ -0,0 +1,33 @@
...
+            RAISE INFO 'Moving link: [%]', link_contract.id;
+
+            -- Move link record from links2 table
+            WITH moved_rows AS (
+                DELETE FROM links2 WHERE id=link_contract.id RETURNING *
+            ) INSERT INTO archive_links2 SELECT * FROM moved_rows;
+
+            -- Move link record from cards table
+            WITH moved_rows AS (
+                DELETE FROM cards WHERE id=link_contract.id RETURNING *
```
Page-, line 24: Am I right in reading this as that contracts are duplicated between `links2` and `cards` with the same id?

sql/archive.sql
```diff
@@ -0,0 +1,33 @@
...
+    contract record;
+    link_contract record;
+BEGIN
+    FOR contract IN SELECT * FROM cards WHERE (type='execute@1.0.0' OR type='session@1.0.0' OR type='external-event@1.0.0' OR active=false)
+        AND created_at < to_char(CURRENT_DATE - INTERVAL '3 months', 'YYYY-MM-DD')::date
+    LOOP
+        RAISE INFO 'Moving contract: [%] [%]', contract.type, contract.id;
+        FOR link_contract IN SELECT * FROM links2 WHERE fromid=contract.id OR toid=contract.id
+        LOOP
+            RAISE INFO 'Moving link: [%]', link_contract.id;
```
Page-, line 15: Whilst looping through 1-by-1 is useful for being able to log each one but in terms of performance it's going to be much slower than writing SQL query/queries that work on the full batch (although precisely what that query/queries will look like depends on the other question)

sql/archive-tables.sql
```diff
@@ -0,0 +1,40 @@
...
+    requires jsonb[] NOT NULL,
+    capabilities jsonb[] NOT NULL,
+    data jsonb NOT NULL,
+    linked_at jsonb NOT NULL,
+    created_at timestamp with time zone NOT NULL,
+    updated_at timestamp with time zone,
+    CONSTRAINT version_positive CHECK (((version_major >= 0) AND (version_minor >= 0) AND (version_patch >= 0)))
+);
+ALTER TABLE ONLY public.archive_cards ADD CONSTRAINT archive_cards_pkey PRIMARY KEY (id);
+ALTER TABLE ONLY public.archive_cards ADD CONSTRAINT archive_cards_slug_version_key UNIQUE (slug, version_major, version_minor, version_patch, version_prerelease, version_build);
```
Page-, line 28: You should be able to create the archive tables matching the pattern of the existing tables in order to avoid any typos/accidental differences (they can still diverge after the fact, but at least the initial state will be identical). I think
```suggestion
CREATE TABLE IF NOT EXISTS public.archive_cards (
    LIKE cards INCLUDING ALL,
);
```
should work to do that

sql/archive-tables.sql
```diff
@@ -0,0 +1,40 @@
...
+CREATE TABLE IF NOT EXISTS public.archive_links2 (
+    id uuid NOT NULL,
+    forward boolean NOT NULL,
+    fromid uuid NOT NULL,
+    name integer NOT NULL,
+    toid uuid NOT NULL
+);
+ALTER TABLE ONLY public.archive_links2 ADD CONSTRAINT archive_links2_pkey PRIMARY KEY (id, forward);
+CREATE INDEX idx_archive_links2_fromid_name_toid ON public.archive_links2 USING btree (fromid, name, toid);
+CREATE INDEX idx_archive_links2_toid_name_fromid ON public.archive_links2 USING btree (toid, name, fromid);
```
Page-, line 40: Same thing, I expect
```suggestion
CREATE TABLE IF NOT EXISTS public.archive_links2 (
    LIKE links2 INCLUDING ALL,
);
```
to clone the links2 table

sql/archive.sql
joshbwlng: Yes, I believe this is so because querying on links using the `cards` table is too slow, so a separate "meta" table specifically for `link` contracts was created.

sql/archive-tables.sql
joshbwlng: Cool, didn't know about that 👍

sql/archive.sql
joshbwlng: Good point. I was mainly using those output statements for debugging purposes, but the function should be optimized to work on our production table with ~155,000,000 records 😓

sql/archive-tables.sql
joshbwlng: Updated in new commit

sql/archive-tables.sql
joshbwlng: Updated in new commit

sql/archive.sql
joshbwlng: @Page- Updated the logic to be bit more efficient and not loop one at a time

sql/archive.sql
```diff
@@ -0,0 +1,42 @@
...
+-- Move old contracts to archive tables
+DO $$
+DECLARE
+    contract_ids UUID[];
+    link_ids UUID[];
+BEGIN
+    LOOP
+        -- Select contracts we no longer need in the main table
+        SELECT array_agg(id) INTO contract_ids FROM cards WHERE (type='execute@1.0.0' OR type='session@1.0.0' OR type='external-event@1.0.0' OR active=false)
+            AND created_at < to_char(CURRENT_DATE - INTERVAL '3 months', 'YYYY-MM-DD')::date LIMIT 1000;
```
joshbwlng, line 12: Should we try and `SELECT FOR UPDATE SKIP LOCKED`? I don't think it will really be an issue for archiving old unused contracts, and also `FOR UPDATE is not allowed with aggregate functions`. Thought I'd bring up this concern though just in case.

ramirogm: @joshbwlng 
After some tests, I think the `LIMIT` clause would need to be applied to a subquery.

Example, compare this alternative:

```
PGPASSWORD=$POSTGRES_PASSWORD  psql --host=$POSTGRES_HOST -U $POSTGRES_USER  -d jellyfish  << EOF

select now();

SELECT array_agg(id) 
FROM (
  select *
  from 
  cards WHERE (type='execute@1.0.0' OR type='session@1.0.0' OR type='external-event@1.0.0' OR active=false)
            AND created_at < to_char(CURRENT_DATE - INTERVAL '3 months', 'YYYY-MM-DD')::date LIMIT 1000

) as subset
;

select now();

EOF
```
which runs in 10ms and has:

```
                                                                    QUERY PLAN                                                                     
---------------------------------------------------------------------------------------------------------------------------------------------------
 Aggregate  (cost=500.95..500.96 rows=1 width=32)
   ->  Limit  (cost=0.58..488.45 rows=1000 width=945)
         ->  Index Scan using created_at_cards_idx on cards  (cost=0.58..16878261.27 rows=34596261 width=945)
               Index Cond: (created_at < (to_char((CURRENT_DATE - '3 mons'::interval), 'YYYY-MM-DD'::text))::date)
               Filter: ((type = 'execute@1.0.0'::text) OR (type = 'session@1.0.0'::text) OR (type = 'external-event@1.0.0'::text) OR (NOT active))
```


**vs**

```
PGPASSWORD=$POSTGRES_PASSWORD  psql --host=$POSTGRES_HOST -U $POSTGRES_USER  -d jellyfish  << EOF

select now();

SELECT array_agg(id)
-- INTO contract_ids 
FROM cards WHERE (type='execute@1.0.0' OR type='session@1.0.0' OR type='external-event@1.0.0' OR active=false)
AND created_at < to_char(CURRENT_DATE - INTERVAL '3 months', 'YYYY-MM-DD')::date LIMIT 1000;

select now();

EOF
```

which I have to cancel and has much higher cost

```
                                                                    QUERY PLAN                                                                     
---------------------------------------------------------------------------------------------------------------------------------------------------
 Limit  (cost=16964772.53..16964772.54 rows=1 width=32)
   ->  Aggregate  (cost=16964772.53..16964772.54 rows=1 width=32)
         ->  Index Scan using created_at_cards_idx on cards  (cost=0.58..16878281.83 rows=34596280 width=16)
               Index Cond: (created_at < (to_char((CURRENT_DATE - '3 mons'::interval), 'YYYY-MM-DD'::text))::date)
               Filter: ((type = 'execute@1.0.0'::text) OR (type = 'session@1.0.0'::text) OR (type = 'external-event@1.0.0'::text) OR (NOT active))
```

ramirogm: Unless the plpgsql `select into` does something different ( still testing )

ramirogm: Also confirmed with:

```
SET statement_timeout=0;

-- Move old contracts to archive tables
DO $$
DECLARE
    contract_ids UUID[];
    link_ids UUID[];
BEGIN
  PERFORM now();
  
  -- Select contracts we no longer need in the main table
  PERFORM array_agg(id) 
  FROM cards 
  WHERE 1=1 
    AND (type='execute@1.0.0' OR type='session@1.0.0' OR type='external-event@1.0.0' OR active=false)
    AND created_at < to_char(CURRENT_DATE - INTERVAL '3 months', 'YYYY-MM-DD')::date 
  LIMIT 1000;

  PERFORM now();

END;
$$ LANGUAGE plpgsql;



-------------------------------


DO $$
DECLARE
    contract_ids UUID[];
    link_ids UUID[];
BEGIN
  PERFORM now();
  
  -- Select contracts we no longer need in the main table
  PERFORM array_agg(id) 
  FROM ( 
    SELECT *
    FROM
      cards 
    WHERE 1=1 
      AND (type='execute@1.0.0' OR type='session@1.0.0' OR type='external-event@1.0.0' OR active=false)
      AND created_at < to_char(CURRENT_DATE - INTERVAL '3 months', 'YYYY-MM-DD')::date 
    LIMIT 1000    
      ) as subset;

  PERFORM now();

END;
$$ LANGUAGE plpgsql;


```

@joshbwlng

joshbwlng: Thanks @ramirogm , updated my PR with your suggestions :+1:

sql/archive.sql
```diff
@@ -0,0 +1,48 @@
...
+            RAISE INFO 'No more matching contracts found, quitting...';
+            EXIT;
+        END IF;
+        RAISE INFO 'Archiving % contracts...', array_length(contract_ids, 1);
+
+        -- Select links for matching contracts
+        SELECT array_agg(id) INTO primary_link_ids FROM links2 WHERE fromid = ANY(contract_ids) OR toid = ANY(contract_ids);
+        SELECT array_agg(id) INTO secondary_link_ids FROM links2 WHERE fromid = ANY(primary_link_ids) OR toid = ANY(primary_link_ids);
+        link_ids = array_cat(primary_link_ids, secondary_link_ids);
+        RAISE INFO 'Archiving % links...', array_length(link_ids, 1);
```
joshbwlng, line 27: @ramirogm @Page- Tested this function on a very small subset of session contracts from 2018/2019 and ran into an issue. Found that there are links linked to other links, which we need to account for. From what I can tell, this happens when, for example, `contract-a` is manually linked `contract-b` by a user. This action results in a `create` contract that is linked by `link-a` to `link-b` which in turn is linking `contract-a` to `contract-b`. A bit confusing, but I think my SQL accounts for this. I also don't think there will be a third layer to this, but won't really know until I run the SQL. Looking for a sanity check!

sql/archive.sql
Ereski: @Page- @joshbwlng just to expand a bit more on this. `link` contracts define a bidirectional edge even though it has a `from` and `to` field. But links also have a name and an inverse name like `has attachment` and `is attached to`. So the query for a link is:

```sql
(link.data.from.id = <parent ID> AND link.name = <name>) OR
(link.data.to.id = <parent ID> AND link.data.inverseName = <name>)
```

Even with proper indexes, PG does not like that. It always uses a bitmap scan, which is kinda horrible here as contracts are large, there's toast and millions of contracts. Index or not, there's still a lot of IO required.

With a separate `links` table, the situation is better because rows are much smaller. Only some fields are duplicated from the full `link` contract. But i doesn't fundamentally fix the issue of bitmap scans. Each row still denotes a bidirectional link so the query above doesn't change much.

With `links2`, each row denotes a single direction of the bidirectional edge. Each `link` contract has *two* rows in the `links2` table associated with it. The query becomes a much simpler:

```sql
links2.from = <parent ID> AND links2.name = <name>
```

PG is happy to do an index-only search and each row was also shrunk in this new version. The improvements in query time and resource usage are dramatic.

sql/generate.sql
```diff
@@ -0,0 +1,308 @@
...
+        ('{"actor": "' || uuid_in(md5(random()::text || clock_timestamp()::text)::cstring) || '", "epoch": ' || extract(EPOCH FROM (NOW() + (random() * (interval '90 days')) + '30 days'))::integer || ', "input": {"id": "' || uuid_in(md5(random()::text || clock_timestamp()::text)::cstring) || '"}, "action": "action-create-card@1.0.0", "context": {"id": "REQUEST-28.0.3-' || uuid_in(md5(random()::text || clock_timestamp()::text)::cstring) || '", "api": "SERVER-28.0.3-localhost-' || uuid_in(md5(random()::text || clock_timestamp()::text)::cstring) || '"}, "arguments": {"reason": null, "properties": {"data": {"actor": "' || uuid_in(md5(random()::text || clock_timestamp()::text)::cstring) || '", "expiration": "' || (NOW() + (random() * (interval '90 days')) + '30 days')::text || '"}, "slug": "session-ui-user-johndoe-' || uuid_in(md5(random()::text || clock_timestamp()::text)::cstring) || '", "linked_at": {}}}, "timestamp": "' || (NOW() + (random() * (interval '90 days')) + '30 days')::text || '"}')::json,
+        '{}',
+        (NOW() + (random() * (interval '90 days')) + '30 days')
+    FROM generate_series(1, 1000000);
+
+    -- Create exactly one execute card for each action-request
```
joshbwlng, line 209: Obviously shouldn't merge this as-is since `execute` contracts no longer exist on clean systems. Keeping it here for now to generate data necessary for testing the archive SQL script.

sql/archive.sql
joshbwlng: @Ereski Do you think we can improve any of the link queries I have currently in the archive script? In this case, I'm thinking we can just get all links for the contracts we're archiving, and not worry about the name of the link.

PR #9368, opened 2022-09-30
UI: Add ability to hide dates with uiSchema

joshbwlng:
Change-type: patch

***

Add ability to hide `created_at` and `updated_at` fields for contract snippets when these two fields are set to `null` in a type contracts `uiSchema` configuration.


joshbwlng:
@balena-ci rebase

joshbwlng:
@balena-ci I self certify!

PR #9390, opened 2022-10-11
Always destroy EC2 test instance

ab77:
Change-type: patch

***

test suite!


ab77:
@balena-ci I self-certify!

PR #9406, opened 2022-10-18
Backend: Handle preflight requests for /hooks

joshbwlng:
Change-type: patch

---

Need to return a `204` response for `OPTIONS` requests before the request body is validated through `isEventValid` methods. These checks throw a 500 because there is no payload to `JSON.parse()` for `OPTIONS` requests. The subsequent request after `OPTIONS` is then processed normally.


joshbwlng:
@balena-ci I self certify!

PR #9409, opened 2022-10-21
Remove transformer support

joshbwlng:
Transformers are no longer being used/managed in Jellyfish. Removing
support for transfomers as the code in jellyish-worker needed to
continue support presents a processing cost for each contract
processed by the worker.

Change-type: major


joshbwlng:
@balena-ci I self certify!

PR #9457, opened 2023-01-20
Enable minification of the UI bundle, dropping the size from ~12.1MB to ~4.6MB

Page-:
Change-type: patch

***

test suite!


Page-:
A huge amount of the minified size is due to plotly.js, but stripping it out is more invasive than I feel comfortable doing

JSReds (approved):

Page-:
image

PR #9503, opened 2023-04-04
Update plugin-balena-api to 6.8.4

joshbwlng:
Updates jsonwebtoken to v9 to address security issue

Change-type: patch

***

test suite!


joshbwlng:
lgtm

